PR #447, opened 2025-12-28
fix(frontend): handle missing groups:read permission gracefully

### LINKED ISSUE #446: bug: /students/search fails for users without groups:read permission
## Bug Description
When a user with the "user" role (not admin/teacher) visits `/students/search`, the page fails to load with a 500 error.

## Steps to Reproduce
1. Create a new user via invitation with role "user" and position "OGS-Büro"
2. Log in as that user
3. Navigate to `/students/search`
4. Page shows error / spinner indefinitely

## Error Message
```
API error: 500 ""
at fetchWithRetry (src/lib/api-helpers.ts:480:13)
at async Object.getGroups (src/lib/api.ts:918:36)
at async SearchPageContent.useEffect.loadInitialData (src/app/students/search/page.tsx:175:31)
```

## Root Cause Analysis
1. `/students/search/page.tsx` line 175 calls `groupService.getGroups()` to load ALL groups for the filter dropdown
2. Backend `/api/groups` returns **403 Forbidden** for users without `groups:read` permission
3. The outer catch block (lines 207-209) logs the error but doesn't set `groupsLoaded = true`
4. Since `groupsLoaded` never becomes `true`, the student search at line 217 never triggers
5. Page gets stuck in loading state

## Backend Logs Confirmation
```
2025/12/28 11:38:02 "GET http://server:8080/api/groups HTTP/1.1" - 403 23B
2025/12/28 11:38:03 "GET http://server:8080/api/groups HTTP/1.1" - 403 23B
```

### Option A: Handle error gracefully in search page
Set `groupsLoaded(true)` in the outer catch block so the page continues without the full groups filter:
```typescript
} catch (error) {
  console.error("Error loading groups:", error);
  setGroups([]); // Use empty groups list
  setGroupsLoaded(true); // Allow page to continue
}
```

### Option B: Use user's accessible groups instead
Replace `getGroups()` with an endpoint that returns only groups the user has access to (their assigned groups + supervised groups).

### Option C: Check permissions before calling
Only call `getGroups()` if the user has the `groups:read` permission.

## Environment
- User role: "user"
- User position: "OGS-Büro" 
- Email: fluettgenau@yahoo.de

## Labels
bug, frontend, permissions

fl0r14n28:
## Summary

Fixes #446 - Users without `groups:read` permission could not access `/students/search` because the groups API returned 403, causing the page to fail with a 500 error.

### Root Causes

1. **Missing `await`** in `route-wrapper.ts`: `executeWithRetry` was returned without `await`, so rejected Promises bypassed the try-catch block → 500 instead of 403
2. **Client-side throwing**: `fetchWithRetry` threw errors for all non-ok responses, triggering Next.js error overlay even when caught
3. **Page loading logic**: `groupsLoaded` state was never set to `true` when `getGroups()` failed, causing the page to hang

### Changes

| File | Change |
|------|--------|
| `route-wrapper.ts` | Add `await` before `executeWithRetry` calls |
| `api-helpers.ts` | Return `null` instead of throwing for 4xx errors in `fetchWithRetry` |
| `api.ts` | Return empty array in `getGroups` when response is null |
| `page.tsx` | Make groups loading non-fatal, always set `groupsLoaded(true)` |

## Test Plan

- [x] Login with a user that has role "user" (no `groups:read` permission)
- [x] Navigate to `/students/search`
- [x] Verify: Page loads without error overlay
- [x] Verify: Groups dropdown is empty but functional
- [x] Verify: Student search works normally

### DIFF
--- a/frontend/src/lib/api-helpers.ts
+++ b/frontend/src/lib/api-helpers.ts
@@ -231,8 +231,8 @@ export async function apiDelete<T>(
 export function handleApiError(error: unknown): NextResponse<ApiErrorResponse> {
   // If it's an Error with a specific status code pattern, extract it
   if (error instanceof Error) {
-    // Match both "API error: 403" and "API error (403):" formats
-    const regex = /API error[:\s(]+(\d+)/;
+    // Match both "API error: 403" and "API error (403):" formats (exactly 3 digits)
+    const regex = /API error[:\s(]+(\d{3})/;
     const match = regex.exec(error.message);
 
     if (match?.[1]) {
@@ -484,13 +484,14 @@ export async function fetchWithRetry<T>(
 
   if (!response.ok) {
     const errorText = await response.text();
-    // 4xx errors (client errors like 403 Forbidden) are expected - return null without throwing
-    // to avoid triggering Next.js error overlay in development
-    if (response.status >= 400 && response.status < 500) {
-      console.warn(`API error: ${response.status}`, errorText);
+    // Only 401/403 are expected "access denied" scenarios - return null for graceful handling
+    // Other 4xx errors (400 Bad Request, 404 Not Found) indicate bugs and should throw
+    const accessDeniedStatuses = [401, 403];
+    if (accessDeniedStatuses.includes(response.status)) {
+      console.warn(`API access denied: ${response.status}`, errorText);
       return { response: null, data: null };
     }
-    // 5xx errors are unexpected server errors - throw to trigger error handling
+    // All other errors (4xx bugs, 5xx server errors) should throw
     console.error(`API error: ${response.status}`, errorText);
     throw new Error(`API error: ${response.status}`);
   }


yungweng (changes requested):
PR #447 Code Review: Handle Missing groups:read Permission Gracefully

  Overview

  This PR fixes a usability issue where users without groups:read permission were getting 500 errors when accessing /students/search. The fix implements graceful degradation so the page loads with an empty groups dropdown instead of failing.

  ---
  ✅ What's Good

  1. The missing await fix is correct (route-wrapper.ts:252, 283)

  // Before (broken): Promise rejection bypasses try-catch
  return executeWithRetry(session.user.token, executeHandler, ...);

  // After (correct): Properly caught by try-catch
  return await executeWithRetry(session.user.token, executeHandler, ...);

  ★ Insight ─────────────────────────────────────
  When a function returns a Promise without await, any rejection happens after the function exits the try-catch block. This is a common async/await pitfall—TypeScript won't catch it because both patterns are type-valid.
  ─────────────────────────────────────────────────

  2. Page restructuring is clean (page.tsx)
  - Groups loading is now non-fatal with proper warning logging
  - setGroupsLoaded(true) is guaranteed to execute (moved outside try-catch)
  - Reduces nested try-catch from 2 levels to 1

  3. Conditional logging based on status (api-helpers.ts:235-242)
  - 5xx → console.error (unexpected server errors)
  - 4xx → console.warn (expected client errors)
  - Reduces noise in Next.js development overlay

  ---
  ⚠️ Concerns & Suggestions

  1. Silent 4xx handling may mask bugs (api-helpers.ts:487-492)

  // Current: ALL 4xx errors return null
  if (response.status >= 400 && response.status < 500) {
    console.warn(`API error: ${response.status}`, errorText);
    return { response: null, data: null };
  }

  Problem: This treats 403 Forbidden (permission denied) the same as 400 Bad Request (malformed request) or 404 Not Found (resource doesn't exist). A 400 likely indicates a frontend bug that should throw.

  Suggestion: Consider a more targeted approach:

  // Only gracefully handle expected "access denied" scenarios
  const gracefulStatuses = [401, 403]; // Auth/permission issues
  if (gracefulStatuses.includes(response.status)) {
    console.warn(`API access denied: ${response.status}`, errorText);
    return { response: null, data: null };
  }
  // Other 4xx errors indicate bugs - throw
  console.error(`API error: ${response.status}`, errorText);
  throw new Error(`API error: ${response.status}`);

  2. Loss of error context in api.ts (api.ts:927-930)

  // Current: Can't distinguish auth failure from permission denied from empty result
  if (response === null || data === null) {
    return [];
  }

  This works for the immediate use case, but callers have no way to know why the array is empty:
  - User has no groups (legitimate empty result)
  - User lacks permission (403)
  - User's token is invalid (401)

  For future consideration: If you need this distinction later, consider returning a discriminated union or adding a status field.

  3. The regex change is fragile (api-helpers.ts:231)

  // Before: Specific pattern
  const regex = /API error \((\d+)\):/;

  // After: Looser pattern
  const regex = /API error[:\s(]+(\d+)/;

  This now matches:
  - ✅ "API error: 403" (new throw format)
  - ✅ "API error (403):" (old format)
  - ⚠️ "API error 500" (unintended?)

  Suggestion: Be explicit about supported formats:

  const regex = /API error[:\s(]+(\d{3})/; // Exactly 3 digits
  // or
  const regex = /API error(?:: (\d+)|(?:\((\d+)\)))/; // Match both explicitly

  ---
  📋 Summary

  | Aspect         | Rating        | Notes                                         |
  |----------------|---------------|-----------------------------------------------|
  | Correctness    | ✅ Good       | Fixes the root cause properly                 |
  | Code Quality   | ✅ Good       | Clean restructuring, good separation          |
  | Error Handling | ⚠️ Acceptable | Works for this case, but may mask future bugs |
  | Test Coverage  | ✅ Good       | Manual test plan covers the scenario          |
  | Risk           | 🟡 Low-Medium | Broad 4xx handling could hide issues          |

  ---
  Recommendation

  Approve with minor suggestions. The core fix is correct and the approach is sound. The concerns are about long-term maintainability rather than immediate correctness.

  If you want to address the 4xx concern before merge:
  // In api-helpers.ts, be more selective:
  const permissionDeniedStatuses = [401, 403];
  if (permissionDeniedStatuses.includes(response.status)) {
    // ... graceful handling
  }

  Otherwise, the PR is good to merge as-is for the immediate fix, with the understanding that the broader 4xx handling is a tradeoff for simpler error handling.

fl0r14n28 (commented):
Thanks for the thorough review! I've addressed your feedback in commit a3a62d74:

### ✅ Implemented

**1. Selective 4xx handling (api-helpers.ts:489-492)**
```typescript
const accessDeniedStatuses = [401, 403];
if (accessDeniedStatuses.includes(response.status)) {
  console.warn(`API access denied: ${response.status}`, errorText);
  return { response: null, data: null };
}
// All other errors (400, 404, 5xx) throw
```

**2. Regex precision (api-helpers.ts:235)**
```typescript
const regex = /API error[:\s(]+(\d{3})/; // Exactly 3 digits
```

### 📝 Acknowledged (Future Improvement)

**Loss of error context** - Noted for future consideration. For this use case (groups filter in student search), empty array is acceptable. If we need distinction between auth/permission/empty later, we can introduce a Result type.

---

Ready for re-review!

yungweng (approved):
## Re-Review: ✅ Approved

All feedback from the initial review has been addressed correctly in commit `a3a62d7`:

### ✅ Verified Changes

**1. Selective 4xx Handling (api-helpers.ts:487-492)**
```typescript
const accessDeniedStatuses = [401, 403];
if (accessDeniedStatuses.includes(response.status)) {
  console.warn(\`API access denied: ${response.status}\`, errorText);
  return { response: null, data: null };
}
// All other errors (400, 404, 5xx) throw
```
✅ Now only 401/403 return null gracefully—400/404 bugs will throw as expected.

**2. Regex Precision (api-helpers.ts:235)**
```typescript
const regex = /API error[:\s(]+(\d{3})/; // Exactly 3 digits
```
✅ `\d{3}` ensures only valid HTTP status codes match.

**3. Conditional Logging (api-helpers.ts:240-245)**
✅ 5xx → `console.error`, 4xx → `console.warn` — reduces Next.js error overlay noise.

**4. Await Fixes (route-wrapper.ts:252, 283)**
✅ `return await executeWithRetry(...)` — Promise rejections now caught by try-catch.

### 📋 Summary

| Check | Status |
|-------|--------|
| 401/403 selective handling | ✅ |
| Regex uses \d{3} | ✅ |
| Conditional logging | ✅ |
| Await before executeWithRetry | ✅ |
| Page loads with empty groups | ✅ |

Ready to merge! 🚀

PR #581, opened 2026-01-09
ci: remove unused deps job from main workflow

yungweng:
## Summary
- Removes the unused `deps` job from main workflow
- Removes the orphaned `dependencies.yml` workflow file
- Removes the unused `setup-go-and-deps` composite action

## Why?

The `deps` job was providing no benefit because:

1. **Each downstream job has its own setup steps** - `lint` and `test` jobs both have their own `setup-go` and `setup-node` with caching enabled
2. **GitHub Actions cache is shared via cache keys** - The cache is shared across workflow runs, not from the `deps` job within the same run
3. **Verified via workflow logs** - All jobs were hitting cache from previous runs, not from `deps`

## Impact

- **Saves ~17s** per workflow run (runner startup + redundant setup)
- **Allows lint and test to start immediately** in parallel instead of waiting for `deps`
- **Simplifies workflow** by removing unnecessary indirection

## Verification

Analyzed workflow run `20864824495` logs:
```
deps / deps:     "Cache hit occurred on primary key..., not saving cache"
lint / backend:  "Cache hit for: setup-go-Linux-x64-ubuntu24-go-1.24.1-..."  (same key!)
test / coverage: "Cache hit for: setup-go-Linux-x64-ubuntu24-go-1.24.1-..."  (same key!)
```

All jobs independently restore from GitHub's shared cache system.


yungweng (commented):
lgtm;r

PR #589, opened 2026-01-11
fix(modal): resolve scroll reset and unify delete confirmation (#587)

fl0r14n28:
## Summary

Fixes modal-related issues including scroll reset, blur effects, and inconsistent delete confirmation patterns.

- **Scroll reset fix**: Prevent page from scrolling to top when opening modals
- **Blur overlay**: Move to BackgroundWrapper so login page modals also have blur effect
- **React StrictMode**: Use refs in modal effects to prevent spurious blur flicker
- **Delete confirmation**: Unify all database pages to use page-level ConfirmationModal pattern

### Modal Infrastructure
- `modal.tsx`: Use refs for openModal/closeModal to avoid effect re-runs
- `form-modal.tsx`: Same ref-based pattern
- `modal-context.tsx`: Add stable fallback object
- `background-wrapper.tsx`: Global blur overlay for all pages
- `responsive-layout.tsx`: Remove duplicate blur overlay

### Database Pages (unified delete pattern)
All pages now use consistent ConfirmationModal at page level:
- `database/students/page.tsx`
- `database/teachers/page.tsx`
- `database/rooms/page.tsx`
- `database/activities/page.tsx`
- `database/groups/page.tsx`
- `database/devices/page.tsx`
- `database/permissions/page.tsx`
- `database/roles/page.tsx`

### Detail Modals
Added `onDeleteClick` prop and removed `InlineDeleteConfirmation`:
- `student-detail-modal.tsx`
- `teacher-detail-modal.tsx`
- `room-detail-modal.tsx`
- `activity-detail-modal.tsx`
- `group-detail-modal.tsx`
- `device-detail-modal.tsx`
- `permission-detail-modal.tsx`
- `role-detail-modal.tsx`

## Test plan

- [x] Open modal on any page - page should not scroll to top
- [x] Login page: "Passwort vergessen" and "Hilfe" modals have blur background
- [x] All database pages: Click delete → detail closes → confirmation opens
- [x] Cancel delete → confirmation closes → detail reopens
- [x] Confirm delete → item deleted, both modals closed

### DIFF
--- a/frontend/src/components/ui/form-modal.test.tsx
+++ b/frontend/src/components/ui/form-modal.test.tsx
@@ -21,7 +21,7 @@ describe("FormModal", () => {
 
   afterEach(() => {
     vi.useRealTimers();
-    document.body.style.overflow = "";
+    document.documentElement.style.overflow = "";
   });
 
   it("should not render when isOpen is false", () => {
@@ -147,7 +147,7 @@ describe("FormModal", () => {
     expect(screen.getByRole("button", { name: "Submit" })).toBeInTheDocument();
   });
 
-  it("should hide body overflow when open", async () => {
+  it("should hide html overflow when open", async () => {
     render(
       <TestWrapper>
         <FormModal isOpen={true} onClose={vi.fn()} title="Test">
@@ -160,10 +160,10 @@ describe("FormModal", () => {
       vi.advanceTimersByTime(20);
     });
 
-    expect(document.body.style.overflow).toBe("hidden");
+    expect(document.documentElement.style.overflow).toBe("hidden");
   });
 
-  it("should restore body overflow when closed", async () => {
+  it("should restore html overflow when closed", async () => {
     const { rerender } = render(
       <TestWrapper>
         <FormModal isOpen={true} onClose={vi.fn()} title="Test">
@@ -188,7 +188,7 @@ describe("FormModal", () => {
       vi.advanceTimersByTime(20);
     });
 
-    expect(document.body.style.overflow).toBe("");
+    expect(document.documentElement.style.overflow).toBe("");
   });
 
   describe("size prop", () => {
--- a/frontend/src/components/ui/form-modal.tsx
+++ b/frontend/src/components/ui/form-modal.tsx
@@ -4,6 +4,7 @@ import { useEffect, useCallback, useState, useRef } from "react";
 import type { ReactNode } from "react";
 import { createPortal } from "react-dom";
 import { useModal } from "../dashboard/modal-context";
+import { useScrollLock } from "~/hooks/useScrollLock";
 import { dialogAriaProps } from "./modal-utils";
 
 interface FormModalProps {
@@ -36,6 +37,9 @@ export function FormModal({
   const closeModalRef = useRef(closeModal);
   openModalRef.current = openModal;
   closeModalRef.current = closeModal;
+
+  // Use scroll lock hook (handles overflow:hidden and event blocking)
+  useScrollLock(isOpen);
 
   // Map size to max-width classes
   const sizeClasses = {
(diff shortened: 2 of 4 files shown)


yungweng (changes requested):
• Switching scroll locking to event-based blocking leaves the window scrollbar active, allowing the background to scroll underneath open modals. This regression breaks the modal scroll lock behavior, so the patch is not fully correct.

  Review comment:

  - [P2] Scroll lock still allows background scroll via scrollbar — /Users/yonnock/Developer/moto/project-phoenix/frontend/src/hooks/useScrollLock.ts:20-25
    The new useScrollLock stops wheel/touch/key events but deliberately leaves the page scrollbar active (// ... scrollbar stays visible), and there is no overflow hide or handler for scrollbar dragging. With any modal open, dragging the window
    scrollbar still scrolls the underlying page while the modal overlay remains fixed, so the background can move underneath the dialog—a regression from the previous overflow-based lock. Reintroduce overflow hiding or handle scrollbar dragging so the
    page truly stays fixed when isLocked is true.

---

• The updated scroll lock drops the overflow-based locking in favor of event prevention, which no longer blocks scrollbar dragging; users can scroll the background while a modal is open. This regression means the patch is not correct.

  Review comment:

  - [P1] Scroll lock no longer blocks scrollbar dragging — /Users/yonnock/Developer/moto/project-phoenix/frontend/src/hooks/useScrollLock.ts:15-25
    Switching useScrollLock to rely solely on wheel/touch/keyboard event prevention (replacing the previous overflow: hidden lock) leaves the page scrollbar fully operational. With a modal open, a user can still click and drag the browser scrollbar to
    scroll the background content because no CSS lock or scroll restoration is applied, breaking the modal’s isolation. Consider retaining overflow-based locking or otherwise disabling scrollbar interactions while isLocked is true.

---

• The updated scroll lock implementation leaves the scrollbar enabled, allowing background pages to scroll while a modal is
  open, which breaks the intended lock behavior. This regression means the patch is not correct overall.

  Review comment:

  - [P1] Scroll lock no longer prevents scrollbar dragging — /Users/yonnock/Developer/moto/project-phoenix/frontend/src/
    hooks/useScrollLock.ts:18-25
    When a modal is open the hook now only adds a modal-open class and relies on wheel/touch/keyboard event handlers,
    leaving the scrollbar visible. Dragging the scrollbar (or programmatic scrolls) still moves the background because
    overflow is never disabled, so the page can scroll under an open modal despite the supposed lock. This regression occurs
    whenever a modal uses useScrollLock(true) on desktop with a visible scrollbar.

---

Review Summary

  The scroll lock refactoring addresses layout shift issues, but introduces some gaps and inconsistencies that should be addressed.

  1. Scrollbar Drag Escape (P2)

  The new event-based approach blocks wheel, touch, and keyboard scrolling, but scrollbar dragging still scrolls the background. While this is a deliberate trade-off to prevent layout shifts, there are solutions that provide both benefits:

  Option A: Position-fixed approach
  body.modal-open {
    position: fixed;
    top: calc(-1 * var(--scroll-y));
    width: 100%;
    overflow-y: scroll; /* Keep scrollbar visible */
  }

  Option B: Modern CSS
  html { scrollbar-gutter: stable; }
  This prevents layout shifts even with overflow: hidden.

  2. modal-open Class is Dead Code (P2)

  The hook adds modal-open to document.body (frontend/src/hooks/useScrollLock.ts#L21), but no CSS rule targets this class in globals.css. Either implement the CSS or remove the class.

  3. Inconsistent Modal Behavior (P2)

  FormModal still uses the old overflow: hidden approach:

  // form-modal.tsx:79
  document.body.style.overflow = "hidden";

  This means Modal and FormModal have different scroll lock behavior, which is confusing and should be unified.

  ---
  Recommendation: Consider using the position: fixed approach with scroll offset preservation. This gives you:
  - ✅ Complete scroll lock (including scrollbar)
  - ✅ No layout shifts
  - ✅ Consistent behavior across all modals

PR #878, opened 2026-02-08
fix: eliminate JWT refresh dead zone, double-rotation, and redundant signIn

yungweng:
## Summary

- **Fix 1 — Dead zone elimination**: Remove the `now < tokenExpiry` guard from the JWT callback so it handles both pre-expiry and post-expiry refresh. Users returning after token expiry no longer fall through to the racy Axios interceptor path.
- **Fix 2 — No more double-rotation**: `/api/auth/token` and `refreshSessionTokensOnServer()` now delegate to `auth()` instead of making independent backend `/auth/refresh` calls, eliminating double token rotation that wasted refresh tokens.
- **Fix 3 — Remove redundant `signIn`**: After `/api/auth/token` calls `auth()`, the session cookie already contains refreshed tokens. The `signIn("credentials", { internalRefresh })` call in `handleAuthFailure` was triggering a redundant JWT callback cycle — removed it.
- **Fix 4 — Post-expiry error signaling**: When access token is already expired and refresh fails, set `token.error = "RefreshTokenError"` to trigger immediate re-login instead of leaving the user with a dead token that silently fails API calls.
- **Fix 5 — Remove dead server-side retry**: Remove 401 retry in `serverFetchWithRetry` — React `cache()` memoizes `auth()` per-request, so the retry always gets the same stale token (dead code).
- **Stress tests**: 14 scenario-based tests covering idle return (30s, 30min expired), concurrent 401s, both-tokens-expired, network timeout, malformed JSON, no-infinite-loop, missing field edge cases, and post-expiry refresh failures.

## What changed

| File | Change |
|------|--------|
| `server/auth/config.ts` | Remove `now < tokenExpiry` guard (dead zone fix); add `else if (now > tokenExpiry)` error branches |
| `app/api/auth/token/route.ts` | Simplify to delegate to `auth()` |
| `server/auth/token-refresh.ts` | Simplify to delegate to `auth()` |
| `lib/auth-api.ts` | Remove redundant `signIn` from `handleAuthFailure` |
| `lib/api-helpers.ts` | Remove dead 401 retry block in `serverFetchWithRetry` |
| `server/auth/jwt-refresh-scenarios.test.ts` | **NEW** — 14 scenario tests for JWT callback |
| `server/auth/token-refresh.test.ts` | Add failure-then-recovery scenario |
| `lib/auth-api.test.ts` | Update assertions for removed `signIn` |
| `app/api/auth/token/route.test.ts` | Simplified to match new route |
| `server/auth/config.test.ts` | Updated for dead zone fix |

## Test plan

- [x] `pnpm vitest run src/server/auth/jwt-refresh-scenarios.test.ts` — 14 scenario tests pass (2 new: post-expiry failures)
- [x] `pnpm vitest run src/lib/api-helpers.test.ts` — 46 tests pass
- [x] `pnpm vitest run src/lib/auth-api.test.ts` — 29 tests pass (updated)
- [x] `pnpm vitest run src/server/auth/token-refresh.test.ts` — 8 tests pass (1 new)
- [x] `pnpm vitest run src/app/api/auth/token/route.test.ts` — 5 tests pass
- [x] `pnpm vitest run src/server/auth/config.test.ts` — 24 tests pass
- [x] `pnpm vitest run` — all 7244 tests pass, zero regressions
- [x] `pnpm run check` — lint + typecheck, zero warnings

### DIFF
--- a/frontend/src/lib/api-helpers.ts
+++ b/frontend/src/lib/api-helpers.ts
@@ -80,16 +80,7 @@ async function serverFetchWithRetry<T>(
       cache: "no-store", // Prevent Next.js from caching API responses
     });
 
-  let response = await executeRequest(token);
-
-  if (response.status === 401) {
-    const { refreshSessionTokensOnServer } =
-      await import("~/server/auth/token-refresh");
-    const refreshed = await refreshSessionTokensOnServer();
-    if (refreshed?.accessToken) {
-      response = await executeRequest(refreshed.accessToken);
-    }
-  }
+  const response = await executeRequest(token);
 
   if (!response.ok) {
     const errorText = await response.text();
--- a/frontend/src/server/auth/config.ts
+++ b/frontend/src/server/auth/config.ts
@@ -423,6 +423,10 @@ export const authConfig = {
             token.error = undefined;
             token.needsRefresh = undefined;
             logger.info("proactive_token_refresh_succeeded");
+          } else if (now > tokenExpiry) {
+            token.error = "RefreshTokenError";
+            token.needsRefresh = true;
+            logger.warn("token_refresh_failed_post_expiry");
           }
           return token;
         }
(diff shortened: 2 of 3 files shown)


yungweng:
push and pray

PR #937, opened 2026-03-01
fix: use token transfer for cross-subdomain tenant switching

christiank119:
Tenant switching failed on localhost because cookies are scoped to the exact hostname (e.g. school-a.localhost) and cannot be shared across subdomains due to the Public Suffix List. The old flow established the session on the source subdomain, then navigated away — losing it.

New approach: pass JWT tokens via URL hash fragment to a callback page on the target subdomain, which establishes the session there. Hash fragments are never sent to the server, avoiding log/referrer exposure.

### DIFF
--- a/frontend/src/app/[tenant]/auth/tenant-callback/page.test.tsx
+++ b/frontend/src/app/[tenant]/auth/tenant-callback/page.test.tsx
@@ -1,0 +1,298 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
+import { render, screen, waitFor } from "@testing-library/react";
+import "@testing-library/jest-dom/vitest";
+
+// ============================================================================
+// Mocks
+// ============================================================================
+
+const { mockSignIn } = vi.hoisted(() => ({
+  mockSignIn: vi.fn(),
+}));
+
+const { mockMutate } = vi.hoisted(() => ({
+  mockMutate: vi.fn(),
+}));
+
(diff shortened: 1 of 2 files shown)


yungweng:
Hey @christiank119, thanks for digging into this and for the thorough write-up — the investigation into the localhost PSL issue is spot-on, and the test coverage is solid.

That said, I don't think we need these code changes. Here's my reasoning:

**Production doesn't have this problem.** Our `auth/config.ts` already sets `Domain=.moto-app.de` on session cookies when `TENANT_DOMAIN` is set and isn't `localhost`. That means cookies are shared across subdomains — `school-a.moto-app.de` and `school-b.moto-app.de` see the same session cookie. The existing flow (signIn on source → navigate to target) works correctly in production.

**This PR changes the production flow too.** The hash fragment redirect is applied unconditionally — not just for localhost. That means production now gets an extra redirect hop through `/auth/tenant-callback` that wasn't needed before, adding latency and a new page to maintain for no benefit.

**The simpler fix already exists — and you documented it yourself.** Your own comment in `auth/config.ts` says: _"To test seamless switching locally, use `TENANT_DOMAIN=lvh.me`"_. That's a one-line env var change that makes `localhost` behave like production. No code changes needed. We just need to document it in `backend/dev.env.example` and the README.

So I'd like to close this unless you can make a case for why the token transfer approach is needed despite the above. Am I missing something?

PR #1023, opened 2026-03-22
fix: resolve staff ID for CSV import pickup schedule created_by

fl0r14n28:
## Summary

- CSV import used the JWT **account ID** as `created_by` for `student_pickup_schedules`, but the composite FK `fk_pickup_schedules_created_by_tenant` references `users.staff(tenant_id, id)`
- This worked accidentally in seeded tenants (where account/staff IDs happened to align) but crashed in newly provisioned tenants with `SQLSTATE=23503`
- Now resolves account → person → staff **inside the tenant transaction** to ensure RLS compatibility

## Test plan

- [x] CSV import in seeded tenant (demo-school) still works
- [x] CSV import in newly created tenant (Burbach) no longer crashes with FK violation
- [x] Preview endpoint also resolves staff ID correctly
- [x] Verify audit log still records account ID (not staff ID)

### DIFF
--- a/backend/api/import/import_test.go
+++ b/backend/api/import/import_test.go
@@ -363,6 +363,95 @@ func TestImportStudents_WithDuplicateData(t *testing.T) {
 }
 
 // =============================================================================
+// STAFF ID RESOLUTION TESTS
+// =============================================================================
+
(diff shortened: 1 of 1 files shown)


fl0r14n28:
Adressing in new PR.

PR #1025, opened 2026-03-22
fix: address review findings — use PersonService and improve error handling

fl0r14n28:
## Summary

Addresses review findings from #1023:

- **Use PersonService instead of raw repos** — replaced `PersonRepository`/`StaffRepository` injection with `PersonService`, matching the pattern used by all other handlers (`pickup_schedule_handlers.go`, `checkin.go`, `groups_handlers.go`, `sse_helpers.go`). This removes the extra params from `NewResource()` and `base.go`.
- **Preserve actual DB errors** — split `err != nil || obj == nil` checks into separate branches so DB timeouts, RLS violations etc. get wrapped with `%w` instead of being masked as "person not found"
- **Add missing test** — `TestImportStudents_PersonWithoutStaff` covers the import handler's error path (500) for person-without-staff, complementing the existing preview handler test (401)

## Test plan

- [x] All 20 import tests pass locally
- [x] `go build ./...` clean
- [x] `pnpm run check` clean


yungweng:
done!

PR #1094, opened 2026-03-25
fix: student search filters from first character

fl0r14n28:
## Summary
- Removed the `searchTerm.length >= 2` guard in the debounce logic on the student search page
- Search now triggers after any character input (with 300ms debounce), not only from the second character onwards
- Backend already handles single-character queries correctly — this was purely a frontend restriction

## Test plan
- [x] Type a single character in the student search — results should filter immediately
- [x] Verify clearing the search input resets the list
- [ ] Verify multi-character search still works as before


yungweng:
nice one! ich mag die Size

yungweng (approved):

PR #1104, opened 2026-03-25
fix: prevent silent delete failures with pre-deletion checks and user-facing error messages

christiank119:
## Summary

  - Deleting staff, rooms, groups, students, persons, guardians, devices, roles, and permissions previously failed
  silently when database constraints blocked the operation — users saw nothing or a cryptic error
  - Added service-layer pre-deletion dependency checks that return HTTP 409 with clear German error messages (e.g.,
  "Gruppe kann nicht gelöscht werden: Gruppe hat noch zugewiesene Schüler/innen")
  - Added constraint violation safety nets in all 9 delete handlers to catch unexpected FK and NOT NULL violations from
  cascading operations
  - Added error toast feedback to all 8 database pages that previously swallowed delete errors
  - Extracted `getDeleteErrorMessage()` helper and `extractErrorMessage()` to clean up nested JSON error responses from
  the API chain

  ## Changes

  **Backend (9 entities protected):**
  - New sentinel errors: `ErrRoomInUse`, `ErrGroupHasStudents`, `ErrStaffInUse`, `ErrDeviceInUse`, `ErrRoleInUse`,
  `ErrPermissionInUse`, etc.
  - New repo methods: `CountByStaffID`, `CountActiveByStudentID`
  - Pre-deletion checks in services (room → active groups, group → enrolled students, staff → active supervisions)
  - `IsConstraintViolation()` helper catches FK violations + NOT NULL violations from composite FK cascades
  - `ErrorConflictMessage()` helper for German user-facing messages without triggering ST1005 linter
  - All constraint errors return HTTP 409 Conflict instead of 500

  **Frontend (8 database pages fixed):**
  - Added catch blocks + `toastError` to rooms, groups, activities, devices, personal, students, roles, permissions
  - `extractErrorMessage()` digs through nested JSON to find the original backend error text
  - `getDeleteErrorMessage()` shared helper eliminates duplication across all pages
  - Downgraded 4xx logging from `error` to `warn` (expected business responses, not server failures)

  ## Test plan

  - [ ] Try deleting a group with students → toast shows "Gruppe kann nicht gelöscht werden: Gruppe hat noch zugewiesene
   Schüler/innen"
  - [ ] Try deleting a room used by a group → toast shows "Raum kann nicht gelöscht werden: Raum wird noch von Gruppen
  verwendet"
  - [ ] Try deleting a staff member with active supervisions → toast shows German error
  - [ ] Try deleting an entity with no dependencies → deletion succeeds as before
  - [ ] Verify no `console.error` for 409 responses (only `console.warn`)
  - [ ] Run `go test ./...` — all backend tests pass
  - [ ] Run `pnpm run check` — frontend lint + typecheck pass

### DIFF
--- a/backend/api/common/errors_test.go
+++ b/backend/api/common/errors_test.go
@@ -3,6 +3,7 @@ package common_test
 import (
 	"encoding/json"
 	"errors"
+	"fmt"
 	"net/http"
 	"net/http/httptest"
 	"testing"
@@ -376,24 +377,20 @@ func TestErrorGone_ResponseBody(t *testing.T) {
 // =============================================================================
 
 func TestIsConstraintViolation(t *testing.T) {
-	t.Run("detects foreign key violation", func(t *testing.T) {
-		err := errors.New(`ERROR: update or delete on table "staff" violates foreign key constraint "fk_attendance_checked_in_by" (SQLSTATE=23503)`)
+	t.Run("detects foreign key violation string", func(t *testing.T) {
+		err := errors.New(`ERROR: update or delete on table "staff" violates foreign key constraint "fk_attendance_checked_in_by"`)
 		assert.True(t, common.IsConstraintViolation(err))
 	})
 
-	t.Run("detects not-null violation from cascading SET NULL", func(t *testing.T) {
-		err := errors.New(`ERROR: null value in column "tenant_id" of relation "groups" violates not-null constraint (SQLSTATE=23502)`)
+	t.Run("detects not-null violation string from cascading SET NULL", func(t *testing.T) {
+		err := errors.New(`ERROR: null value in column "tenant_id" of relation "groups" violates not-null constraint`)
 		assert.True(t, common.IsConstraintViolation(err))
 	})
 
-	t.Run("detects SQLSTATE 23503 code", func(t *testing.T) {
-		err := errors.New("some error SQLSTATE=23503")
-		assert.True(t, common.IsConstraintViolation(err))
-	})
-
-	t.Run("detects SQLSTATE 23502 code", func(t *testing.T) {
-		err := errors.New("some error SQLSTATE=23502")
-		assert.True(t, common.IsConstraintViolation(err))
+	t.Run("detects wrapped FK violation string", func(t *testing.T) {
+		inner := errors.New(`violates foreign key constraint "fk_test"`)
+		wrapped := fmt.Errorf("database error: %w", inner)
+		assert.True(t, common.IsConstraintViolation(wrapped))
 	})
 
 	t.Run("returns false for nil error", func(t *testing.T) {


yungweng (changes requested):
Hey @christiank119 , thanks for tackling this. Silent delete failures have been a pain point for a while, and the overall approach (service-layer pre-checks + constraint safety nets + frontend toasts) is solid. Nice work.

That said, a few things need to change before we can merge this.

## Must fix

**`IsConstraintViolation` needs typed error checking, not string matching**
`backend/api/common/errors.go:163-172`

Right now this checks `err.Error()` for English substrings like `"violates foreign key constraint"`. That breaks if the driver changes its message format. BUN gives us `pgdriver.Error` with a structured `Field('C')` for the SQLSTATE code. Use `errors.As` instead:

```go
var pgErr *pgdriver.Error
if errors.As(err, &pgErr) {
    code := pgErr.Field('C')
    return code == "23503" || code == "23502"
}
```

You can keep the string matching as a fallback if you want, but the typed check should be the primary path.

## Should fix

**Missing `TenantWhere` in new repo methods**
`backend/database/repositories/active/attendance_repository.go:240` (`CountByStaffID`)
`backend/database/repositories/active/visits.go:617` (`CountActiveByStudentID`)

Every other method in those files includes the `TenantWhere` filter as defense-in-depth. These two new methods skip it. Please add it for consistency, even though RLS provides a backstop.

**Staff deletion business logic belongs in a service method**
`backend/api/staff/api.go:676-692`

The staff handler does the pre-check (`FindActiveByStaffID`) and teacher cleanup directly inside the handler's transaction closure. Rooms and groups correctly delegate this to service methods (`facilities.DeleteRoom`, `education.DeleteGroup`). Staff should follow the same pattern with something like `users.DeleteStaff`. Keeps the handler thin.

**Frontend `delete()` should not toast 500 errors as-is**
`frontend/src/lib/database/service-factory.ts:411-440`

All errors, including 500s, get returned as strings for toast display. A user seeing `"API error: 500 - Internal Server Error"` in a toast is not helpful. For non-409 responses, show a generic message like "Ein unerwarteter Fehler ist aufgetreten" and log the actual error. The old `logger.error` calls from `personal/page.tsx` and `students/page.tsx` were also removed in this PR, so unexpected errors now go unlogged on the page level.

## Nice to have (not blocking)

- The German messages are duplicated between sentinel errors and handler safety nets with slightly different wording. Consider extracting them to constants or at least adding a comment that the difference is intentional (pre-check vs generic fallback).
- The pre-check count queries (`CountByGroupIDs`, `CountActiveByStudentID`) silently swallow errors and fall through to the delete. That is fine as a design choice, but a `warn`-level log when the pre-check itself fails would help with debugging.
- Add a brief comment on the check-then-delete pattern noting that the pre-check is best-effort and the real protection is the DB constraint (for future readers).

Everything else looks good. The test changes are justified, the 409 status codes are correct, the `extractErrorMessage` helper is well done, and the consistent page update pattern is clean.

PR #1138, opened 2026-03-26
feat: tenant visibility control and soft-delete/purge

### LINKED ISSUE #1121: demo löschen -> prod instanzen nutzen
-> Demo Seeder für prod Tenants
-> Löschen von Tenants verfügbar machen
-> weniger sources of truth, weniger dependencies.

@fl0r14n28 wdyt?

theitger:
### Summary

Implements two features that eliminate the need for a separate demo deployment:

1. **Hidden schools** — Operators can hide schools from the public landing page while keeping them accessible via direct link. This allows hosting demo/trial tenants on production without exposing them publicly.
2. **Tenant lifecycle management** — Soft-delete, restore, and hard purge for schools. Operators can deactivate a tenant (reversible), restore it, or permanently delete all tenant data when a customer leaves.

---

#### 1. Landing Page Visibility (`hidden` field)

- New `hidden` boolean column on `platform.schools` (migration `1.15.19`)
- `ListActive` (used by the public tenant selector / landing page) now filters `hidden=true` schools
- Hidden schools remain fully accessible via direct subdomain link — `hidden` only affects the public listing
- `FindActiveByAccountID` (used for login/tenant switching) intentionally includes hidden schools — staff can still log in
- Operator dashboard shows a "Verborgen" badge on hidden school cards
- Edit school modal has a "Sichtbarkeit" toggle (Sichtbar/Verborgen) with explanatory text

#### 2. Soft-Delete

- New `deleted_at` timestamptz column on `platform.schools` (migration `1.15.20`)
- Partial index `idx_schools_not_deleted` for efficient queries on non-deleted schools
- `DELETE /operator/schools/{id}` sets `deleted_at = NOW()` — preserves the `active` field so restore returns the school to its exact pre-deletion state
- Deleted schools are excluded from: `FindBySlug`, `FindBySubdomain`, `ListActive`, `FindActiveByAccountID`
- `FindByOrganizationAndSlug` intentionally includes deleted schools to prevent slug reuse while in trash
- Auth layer (`resolveAccountTenantDefault`) blocks login to deleted or inactive tenants
- Tenant resolution cache is invalidated via `revalidateTag` + `revalidatePath` on soft-delete

#### 3. Restore

- `POST /operator/schools/{id}/restore` clears `deleted_at`
- School returns to its pre-deletion state (active/inactive preserved)
- Tenant cache invalidated so the school becomes accessible again immediately

#### 4. Hard Purge

- `POST /operator/schools/{id}/purge` with `{"confirm": "PURGE"}` body
- Requires the school to be soft-deleted first (two-step safety)
- Calls `platform.purge_tenant()` stored procedure (migration `1.15.21`) which:
  - Dynamically discovers all 58+ tenant-scoped tables via `pg_constraint` (no hardcoded table list to maintain)
  - Topologically sorts tables by FK dependency depth
  - Deletes from deepest (leaf) tables first to avoid FK violations
  - Uses `format('%I.%I', schema, table)` for safe identifier quoting
  - Deletes the school record itself
  - Returns a JSONB summary of tables processed and rows deleted
  - Runs as `SECURITY DEFINER` (bypasses RLS); execution restricted to `phoenix_admin` role
- Audit log is written in a separate transaction before the purge, so the audit trail survives even if the purge fails
- Returns the JSONB summary to the operator for transparency

#### 5. Automated Batch Cleanup

- New CLI command: `cleanup tenants [--older-than 90d] [--dry-run]`
- `PurgeDeletedSchools` service method iterates soft-deleted schools older than the retention period
- Each school is purged in its own transaction — partial failures don't block others
- Dry-run mode lists candidates without deleting

#### 6. Operator Dashboard UI

- **Active schools view**: School cards now have a "Löschen" button and "Verborgen" badge
- **Trash view**: Toggle via "Papierkorb (N)" button; shows deleted schools with relative deletion time, "Wiederherstellen" and "Endgültig löschen" buttons
- **Confirmation modals**: Soft-delete (mild warning), restore (informational), purge (strong warning listing what will be destroyed)
- All operations show inline error messages on failure

---

### Audit Trail

All operations are logged to `platform.operator_audit_logs` with:
- `soft_delete`, `restore`, `purge` action types
- Operator ID and client IP
- School metadata (name, slug, subdomain, deleted_at for purge)

---

### Migrations

| Version | Description |
|---------|-------------|
| `1.15.19` | Add `hidden BOOLEAN NOT NULL DEFAULT FALSE` to `platform.schools` |
| `1.15.20` | Add `deleted_at TIMESTAMPTZ` + partial index on non-deleted schools |
| `1.15.21` | Create `platform.purge_tenant(BIGINT)` stored procedure |

---

### API Endpoints (new)

| Method | Path | Description |
|--------|------|-------------|
| `DELETE` | `/operator/schools/{id}` | Soft-delete a school |
| `POST` | `/operator/schools/{id}/restore` | Restore a soft-deleted school |
| `POST` | `/operator/schools/{id}/purge` | Permanently delete all tenant data (requires `{"confirm": "PURGE"}`) |

The existing `PUT /operator/schools/{id}` now accepts a `hidden` boolean field.

---

### Test Coverage

- **52+ new tests** across all layers
- Backend: model, error types, handler, service (unit + sqlmock), repository (integration), stored procedure (integration with real DB verifying cross-schema deletion and tenant isolation)
- Frontend: API mapping, edit modal visibility toggle, page hidden badge, revalidation route
- All tests hermetic (IDs in 900+ range, no hardcoded low IDs)

---

### Test Plan

- [ ] Run migrations on test DB and verify all three apply cleanly
- [ ] Create a school, set hidden=true via edit modal, verify it disappears from public landing page but is still accessible via direct subdomain
- [ ] Soft-delete a school, verify it appears in trash view, login is blocked, and tenant selector no longer shows it
- [ ] Restore a soft-deleted school, verify it reappears and login works again
- [ ] Purge a soft-deleted school, verify all tenant data is removed across schemas
- [ ] Run `cleanup tenants --dry-run` and verify output lists expected candidates
- [ ] Verify audit logs contain entries for all soft-delete/restore/purge operations

### DIFF
--- a/backend/database/repositories/platform/school_repository_test.go
+++ b/backend/database/repositories/platform/school_repository_test.go
@@ -249,6 +249,216 @@ func TestSchoolRepository_QueryMethods(t *testing.T) {
 		assert.True(t, foundHidden, "List should include hidden schools for operator management")
 	})
 
+	t.Run("soft delete sets deleted_at", func(t *testing.T) {
+		sdSchool := &platformModels.School{
+			Model:          modelBase.Model{ID: now + 20},
+			OrganizationID: orgA.ID,
+			Name:           fmt.Sprintf("SoftDel School %d", now),
+			Slug:           fmt.Sprintf("softdel-%d", now),
+			Subdomain:      fmt.Sprintf("softdel-%d", now),
+			Active:         true,
+		}
+		require.NoError(t, repo.Create(ctx, sdSchool))
+		t.Cleanup(func() {
+			_, _ = db.ExecContext(ctx, `DELETE FROM platform.schools WHERE id = ?`, sdSchool.ID)
+		})
+
+		err := repo.SoftDelete(ctx, sdSchool.ID)
+		require.NoError(t, err)
+
+		found, err := repo.FindByID(ctx, sdSchool.ID)
+		require.NoError(t, err)
+		require.NotNil(t, found)
+		assert.True(t, found.IsDeleted(), "school should be soft-deleted")
+		assert.True(t, found.Active, "active flag should be preserved")
+	})
+
+	t.Run("soft delete fails on already-deleted school", func(t *testing.T) {
+		sdSchool := &platformModels.School{
+			Model:          modelBase.Model{ID: now + 21},
+			OrganizationID: orgA.ID,
+			Name:           fmt.Sprintf("SoftDel2 School %d", now),
+			Slug:           fmt.Sprintf("softdel2-%d", now),
+			Subdomain:      fmt.Sprintf("softdel2-%d", now),
+			Active:         true,
+		}
+		require.NoError(t, repo.Create(ctx, sdSchool))
+		t.Cleanup(func() {
+			_, _ = db.ExecContext(ctx, `DELETE FROM platform.schools WHERE id = ?`, sdSchool.ID)
+		})
+
+		require.NoError(t, repo.SoftDelete(ctx, sdSchool.ID))
+
+		err := repo.SoftDelete(ctx, sdSchool.ID)
+		require.Error(t, err, "second soft-delete should fail")
+	})
+
+	t.Run("restore clears deleted_at", func(t *testing.T) {
+		rSchool := &platformModels.School{
+			Model:          modelBase.Model{ID: now + 22},
+			OrganizationID: orgA.ID,
+			Name:           fmt.Sprintf("Restore School %d", now),
+			Slug:           fmt.Sprintf("restore-%d", now),
+			Subdomain:      fmt.Sprintf("restore-%d", now),
+			Active:         true,
+		}
+		require.NoError(t, repo.Create(ctx, rSchool))
+		t.Cleanup(func() {
+			_, _ = db.ExecContext(ctx, `DELETE FROM platform.schools WHERE id = ?`, rSchool.ID)
+		})
+
+		require.NoError(t, repo.SoftDelete(ctx, rSchool.ID))
+		err := repo.Restore(ctx, rSchool.ID)
+		require.NoError(t, err)
+
+		found, err := repo.FindByID(ctx, rSchool.ID)
+		require.NoError(t, err)
+		require.NotNil(t, found)
+		assert.False(t, found.IsDeleted(), "school should be restored")
+	})
+
+	t.Run("restore fails on non-deleted school", func(t *testing.T) {
+		err := repo.Restore(ctx, schoolA.ID)
+		require.Error(t, err, "restore on non-deleted school should fail")
+	})
+
+	t.Run("find deleted older than returns matching schools", func(t *testing.T) {
+		fdSchool := &platformModels.School{
+			Model:          modelBase.Model{ID: now + 23},
+			OrganizationID: orgA.ID,
+			Name:           fmt.Sprintf("FindDel School %d", now),
+			Slug:           fmt.Sprintf("finddel-%d", now),
+			Subdomain:      fmt.Sprintf("finddel-%d", now),
+			Active:         true,
+		}
+		require.NoError(t, repo.Create(ctx, fdSchool))
+		t.Cleanup(func() {
+			_, _ = db.ExecContext(ctx, `DELETE FROM platform.schools WHERE id = ?`, fdSchool.ID)
+		})
+
+		require.NoError(t, repo.SoftDelete(ctx, fdSchool.ID))
+
+		// Backdate deleted_at to 100 days ago
+		_, err := db.ExecContext(ctx,
+			`UPDATE platform.schools SET deleted_at = NOW() - INTERVAL '100 days' WHERE id = ?`,
+			fdSchool.ID)
+		require.NoError(t, err)
+
+		// Should find with 90-day window
+		results, err := repo.FindDeletedOlderThan(ctx, 90*24*time.Hour)
+		require.NoError(t, err)
+		found := false
+		for _, s := range results {
+			if s.ID == fdSchool.ID {
+				found = true
+			}
+		}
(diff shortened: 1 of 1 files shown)


yungweng (commented):
## Follow-up: Additional Critical Issues Found

After further review, several functional regressions surfaced that need to be addressed before merge. Adding these to my earlier review.

---

### P1: `ListActive()` breaks scheduler for hidden schools

`backend/database/repositories/platform/school_repository.go:165` — `ListActive()` now filters `hidden = false`, but **the scheduler also uses `ListActive()`** to iterate tenants for daily cleanup, session-end, and break tasks. A hidden-but-active school silently stops receiving all scheduled jobs even though:
- The PR description says hidden only affects the landing page
- Staff can still log in and use the school via direct link

**Fix**: Either add a separate `ListAllActive()` method (excludes deleted, includes hidden) for the scheduler, or move the `hidden` filter to the API/handler layer where tenant-selector responses are built.

---

### P1: Login fallback drops multi-tenant accounts to `tenant_id=0`

`backend/services/auth/auth_login.go:483-490` — `resolveAccountTenantDefault` picks the first `account_tenants` mapping and returns `(0, 0, nil)` if that school is deleted or inactive. But `FindActiveByAccountID` can return **multiple** mappings. An account belonging to schools A (deleted) and B (active) will get `tenant_id=0` instead of falling through to school B.

**Fix**: Loop through the returned mappings and skip deleted/inactive schools instead of bailing on the first one.

---

### P1: `FindBySubdomain` allows subdomain reuse while soft-deleted — breaks restore

`backend/database/repositories/platform/school_repository.go:129` — `FindBySubdomain` filters `deleted_at IS NULL`, but `platform.schools.subdomain` has a global `UNIQUE` constraint. This means:
1. Operator soft-deletes school with subdomain `demo`
2. Another operator creates a new school with subdomain `demo` — `FindBySubdomain` returns nil (deleted school filtered out), so no conflict detected
3. DB-level `INSERT` or `UPDATE` fails on the UNIQUE constraint

Worse: if step 2 somehow succeeds (e.g., via different code path), **restoring the original school becomes impossible** because the subdomain is taken.

**Fix**: `FindBySubdomain` should include deleted schools (same pattern as `FindByOrganizationAndSlug`), or add a comment explaining why the UNIQUE constraint is sufficient protection.

Wait — actually the UNIQUE constraint *will* prevent step 2 at the DB level, so there's no data corruption risk. But the error message will be a generic constraint violation instead of a clear "subdomain already taken (by deleted school)" message. Consider whether this UX is acceptable.

---

### P1: Provisioning flows don't check `deleted_at`

`loadActiveSchool()` in the service layer (used by `InviteSchoolAdmin`, `CreateSchoolAccount`, device creation) only checks `school.Active` — not `school.IsDeleted()`. After soft-deleting an active school, operators can still:
- Create new accounts on the deleted school
- Send admin invitations
- Create/manage IoT devices

This contradicts the expected behavior that a deleted school is unreachable.

**Fix**: Add `if school.IsDeleted() { return &SchoolNotFoundError{} }` to `loadActiveSchool()`.

---

### P2: Purge doesn't invalidate tenant-resolve cache

`frontend/src/app/operator/provisioning/page.tsx:504-507` — `handlePurge` calls `mutateSchools()` but never calls `revalidateTenantCache(purgeTarget.subdomain)`. After purging, the `[tenant]/layout.tsx` cache (5-min TTL on `tenant-${slug}` tag) still resolves the dead subdomain. Soft-delete and restore both call `revalidateTenantCache`; purge should too.

**Fix**: Add `await revalidateTenantCache(purgeTarget.subdomain);` after `mutateSchools()` in `handlePurge`.

---

### P2: Service layer collapses DB errors into SchoolNotFoundError

`backend/services/platform/operator_provisioning_service.go:1190-1192`:
```go
school, err := s.schoolRepo.FindByID(adminCtx, schoolID)
if err != nil || school == nil {
    return &SchoolNotFoundError{SchoolID: schoolID}
}
```

If `FindByID` fails due to a real database error (connection timeout, etc.), it gets mapped to `SchoolNotFoundError` → HTTP 404/409, hiding the actual failure. The same pattern appears in `RestoreSchool` (line 1214) and `PurgeSchool` (line 1239).

**Fix**: Check `err != nil` and `school == nil` separately. Only the nil-school case should produce `SchoolNotFoundError`; other errors should bubble up as-is.

---

### P2: IoT device auth doesn't check `deleted_at`

The IoT auth stack (`services/iot.GetDeviceByAPIKey` + `auth/device/device_auth.go`) never checks `platform.schools.deleted_at`. Kiosks with valid API keys continue sending check-in/attendance traffic after soft-delete. This is arguably acceptable for the 15-min JWT window, but IoT devices use long-lived API keys, not JWTs.

Whether this is a blocker depends on the intended semantics: does soft-delete mean "hide from UI but keep running" or "fully disable the tenant"? If the latter, device auth needs a `deleted_at IS NULL` check. If the former, document this explicitly.

---

### Summary of new findings

| # | Severity | Issue | Scope |
|---|----------|-------|-------|
| 1 | **P1** | `ListActive()` breaks scheduler for hidden schools | Backend |
| 2 | **P1** | Login fallback drops multi-tenant users to `tenant_id=0` | Backend |
| 3 | **P1** | `FindBySubdomain` + UNIQUE constraint = confusing error on reuse | Backend |
| 4 | **P1** | Provisioning flows work on deleted schools | Backend |
| 5 | **P2** | Purge skips tenant cache invalidation | Frontend |
| 6 | **P2** | DB errors collapsed into SchoolNotFoundError | Backend |
| 7 | **P2** | IoT devices remain active after soft-delete | Backend |

Items 1, 2, and 4 are functional regressions that affect existing behavior. Recommend fixing before merge.

PR #1165, opened 2026-03-28
fix: announcement org/tenant targeting

theitger:
## Summary

Adds organization and tenant-level targeting for platform announcements. Operators can now scope announcements to specific organizations, schools (tenants), or combinations — using OR-union semantics (org match OR tenant match OR global).

### Backend

- **SQL safety**: Replaced `fmt.Sprintf` SQL interpolation with `pgdialect.Array` overlap operator (`&&`) for role matching — eliminates injection risk
- **Shared query logic**: Extracted `unreadWhereClause` constant and `buildUnreadArgs()` helper, shared by `GetUnreadForUser` and `CountUnread` — no more duplicated SQL
- **Soft-delete awareness**: `SchoolRepository.CountByIDs` now excludes soft-deleted schools; `GetStats` always joins `platform.schools` with `deleted_at IS NULL` to avoid counting users in deleted tenants
- **Validation**: Added `MaxTargetIDs = 100` cap on targeting arrays; `validateTargetingIDs` deduplicates before counting
- **Migration**: Partial GIN indexes on `target_org_ids` / `target_tenant_ids` — only indexes non-empty arrays since global announcements use equality check

### Frontend

- **Token refresh fix**: Removed `useSession().status` guard from SWR keys — `OperatorAuthGuard` already ensures auth, and the conditional key caused data loss when status briefly became `"loading"` during token refresh
- **Toast notifications**: Added `useToast()` for success/error feedback on save, delete, and publish actions (replaces silent logger-only errors)
- **Orphan pruning**: `useEffect` prunes selected tenant IDs when their parent org is deselected
- **Safe ID parsing**: `parseInt` + `isNaN` filter replaces bare `Number()` conversion

### Tests

- **Repository layer** (7 tests): Global visibility, org-targeted, tenant-targeted, OR-union, roles+org combo, CountUnread, CountByIDs for orgs and schools
- **Service layer** (12 tests): Create/update/delete/publish/unpublish, unread, count, seen, dismissed, stats, view details, targeting validation
- **API layer** (2 tests): GetUnread success + empty roles
- **Soft-delete** (1 test): Verifies accounts in deleted schools excluded from stats
- **Frontend** (9 tests): handleSave create/update with parseInt ID conversion, handleDelete, handlePublish (success + error toasts), orphan tenant pruning

## Test plan

- [ ] `cd backend && go test ./...` — all backend tests pass
- [ ] `cd backend && go test ./test/ -run TestHermeticTestPatterns -v` — hermetic check passes
- [ ] `cd frontend && pnpm run check` — zero warnings
- [ ] `cd frontend && npx vitest run src/app/operator/announcements/` — 37 tests pass
- [ ] Manual: create announcement with org targeting → only users in that org see it
- [ ] Manual: create announcement with tenant targeting → only users in that school see it
- [ ] Manual: global announcement (no targeting) → all users see it
- [ ] Manual: verify toast notifications on save/delete/publish success and failure

### DIFF
--- a/frontend/src/app/operator/announcements/page.test.tsx
+++ b/frontend/src/app/operator/announcements/page.test.tsx
@@ -931,6 +931,61 @@ describe("OperatorAnnouncementsPage", () => {
       ).toBeInTheDocument();
     });
 
+    it("preserves unknown tenant IDs (soft-deleted schools) during pruning", async () => {
+      // Announcement references tenant id "99" which is not in the schools picker
+      const announcementWithUnknownTenant = {
+        ...mockAnnouncement,
+        targetOrgIds: ["1"],
+        targetTenantIds: ["10", "99"],
+      };
+
+      setupSWRWithOrgAndSchools([announcementWithUnknownTenant]);
+      render(<OperatorAnnouncementsPage />);
+
+      // Open edit modal
+      const menuButtons = screen.getAllByLabelText("Menü öffnen");
+      fireEvent.click(menuButtons[0]!);
+
+      await waitFor(() => {
+        fireEvent.click(screen.getByText("Bearbeiten"));
+      });
+
+      await waitFor(() => {
+        expect(screen.getByTestId("modal")).toBeInTheDocument();
+      });
+
+      const modal = screen.getByTestId("modal");
+
+      // Schule A (id=10) should be selected (it's in Org Alpha which is selected)
+      const schuleABtn = Array.from(modal.querySelectorAll("button")).find(
+        (b) => b.textContent?.includes("Schule A"),
+      );
+      expect(schuleABtn?.className).toContain("border-[#83CD2D]");
+
+      // Now deselect Org Alpha — Schule A (id=10) should be pruned,
+      // but unknown id "99" must be preserved since it's not in the picker
+      const orgAlphaBtn = Array.from(modal.querySelectorAll("button")).find(
+        (b) => b.textContent?.includes("Org Alpha"),
+      );
+      fireEvent.click(orgAlphaBtn!);
+
+      // Re-select Org Alpha to see that Schule A is no longer selected
+      const updatedModal = screen.getByTestId("modal");
+      const orgAlphaBtnAgain = Array.from(
+        updatedModal.querySelectorAll("button"),
+      ).find((b) => b.textContent?.includes("Org Alpha"));
+      fireEvent.click(orgAlphaBtnAgain!);
+
+      await waitFor(() => {
+        const finalModal = screen.getByTestId("modal");
+        const updatedSchuleABtn = Array.from(
+          finalModal.querySelectorAll("button"),
+        ).find((b) => b.textContent?.includes("Schule A"));
+        // Schule A was pruned because its parent org was deselected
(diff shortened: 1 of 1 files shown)


yungweng (changes requested):
## Code Review — Changes Requested

Good work overall — SQL parameterization is solid, the shared `unreadWhereClause` + `buildUnreadArgs` cleanup is a nice DRY win, and test coverage is strong. Two issues need fixing before merge, plus a few smaller items.

---

### P1: `CountByIDs` soft-delete filter breaks existing announcements

**`backend/database/repositories/platform/school_repository.go:249-258`**

Adding `deleted_at IS NULL` to `CountByIDs` makes `validateTargetingIDs` reject any announcement that references a soft-deleted school. This is a regression:

- The announcements UI sources its school list from `schoolRepo.List()` / `listSchools()`, which does **not** filter soft-deleted schools — so operators can still select them
- Existing announcements already in the DB may carry deleted school IDs — simply editing and re-saving one of these now fails with "school IDs do not exist"

**Fix options:**
- A) Remove the `deleted_at IS NULL` filter from `CountByIDs` (validation should accept whatever the picker offers)
- B) Also filter soft-deleted schools from the picker data source so the UI and validation are consistent — but you'd still need a migration or fixup for existing announcements that already reference deleted schools

---

### P1: Success toast fires before `mutate()` can reject

**`frontend/src/app/operator/announcements/page.tsx:219-224`**

SWR's bound `mutate()` rejects on revalidation failure by default. Current ordering:

```
await fetch(...)   // succeeds
toast.success(...) // shown
closeModal()
await mutate()     // can throw → falls into catch → error toast
```

The operator sees a success toast followed by an error toast for an announcement that actually saved. In the **create** flow this false negative is likely to trigger a retry → duplicate announcements. Same pattern repeats in delete/publish handlers.

**Fix:** Move the success toast + `closeModal()` after `await mutate()`, or wrap `mutate()` separately so its failure doesn't surface as a save error.

---

### P2: Partial GIN index change won't reach existing deployments

**`backend/database/migrations/001015022_announcement_org_tenant_targeting.go:65-80`**

Editing the historical `1.15.22` migration to use partial indexes (`WHERE target_org_ids != '{}'`) only affects fresh installs. Staging/production that already applied `1.15.22` keep the old full indexes. If the partial indexes matter for performance, add a **new** migration that drops and recreates them.

---

### Minor items (non-blocking)

- **Frontend pruning edge case** (`page.tsx:126`): When all orgs are deselected, `orgIdKey === ""` skips the pruning `useEffect`, leaving orphaned tenant selections. Consider clearing tenant selections when `targetOrgIds` is empty instead of returning early.

- **Hermetic skip list**: `announcement_service_test.go` is mock-based and uses hardcoded IDs (`base.Model{ID: 1}`, etc.). CI passes because the regex doesn't match this pattern, but for consistency add it to `skipPatterns` in `hermetic_verification_test.go`.

- **Soft-deleted schools in `GetUnreadForUser`**: `GetStats` correctly excludes soft-deleted schools, but `GetUnreadForUser` does not — users in deleted schools can still receive targeted announcements. Worth documenting the intended behavior even if you decide this is fine.

---

@theitger happy to re-review once the P1s are addressed. Nice cleanup on the SQL side 👍

PR #1167, opened 2026-03-29
fix: hide invite buttons for non-admin users on personal page

### LINKED ISSUE #1114: Wrap invite button in RoleGuard on personal page
## Context

PR #1112 moved the invitation form from `/invitations` into `/database/personal`. The "Personal hinzufügen" button and modal are rendered outside `RoleGuard`, so non-admin staff can open the invite form. The backend rejects the submission (permissions check), so this is a UX issue, not a security issue — but non-admins see a workflow they can't complete.

## Expected behavior

The "+" button (desktop) and FAB (mobile) that open the invite modal should only be visible to admins, matching the previous behavior where `/invitations` was fully wrapped in `RoleGuard(adminOnly)`.

## Files

- `frontend/src/app/[tenant]/(protected)/database/personal/page.tsx` — wrap the action button and mobile FAB in `RoleGuard variant="adminOnly"`

theitger:
Closes #1114

## Summary

- Gate the desktop "+" action button and mobile FAB on the personal page behind an `isAdmin(session)` check so non-admin staff no longer see invite controls they can't use

## Why `isAdmin()` instead of `<RoleGuard>`

The issue suggested wrapping the buttons in `<RoleGuard variant="adminOnly">`. However, `RoleGuard` renders a `<ForbiddenPage>` component when the user lacks permission — that makes sense for page sections (and is already used for the `PendingInvitationsList` on this page), but for an individual button it would inject a "forbidden" fragment into the layout instead of simply hiding the button.

Using `isAdmin(session)` directly achieves the correct UX: the button is not rendered at all for non-admins. The underlying logic is identical — `RoleGuard` with `variant="adminOnly"` calls the same `isAdmin(session)` function internally.

## Test plan

- [ ] Log in as admin → "+" button (desktop) and FAB (mobile) visible, invite modal works
- [ ] Log in as non-admin staff → neither button renders, no way to open invite modal
- [ ] Verify `PendingInvitationsList` remains hidden for non-admins (existing `RoleGuard`)

### DIFF
--- a/frontend/src/app/[tenant]/(protected)/database/personal/page.test.tsx
+++ b/frontend/src/app/[tenant]/(protected)/database/personal/page.test.tsx
@@ -5,7 +5,10 @@ import TeachersPage from "./page";
 // Mock next-auth/react
 vi.mock("next-auth/react", () => ({
   useSession: vi.fn(() => ({
-    data: { user: { id: "1", token: "test-token" }, expires: "2099-01-01" },
+    data: {
+      user: { id: "1", token: "test-token", isAdmin: true },
+      expires: "2099-01-01",
+    },
     status: "authenticated",
   })),
 }));


yungweng:
@theitger guardest du den Direktlink oder was da los?

PR #1168, opened 2026-03-29
feat: tenant-scoped settings system

yungweng:
> **⚠️ NOT DONE — this PR is a draft for visibility only.**
  > **Feature owner: @christiank119 christiank119**

  ---

  ## Summary

  Implements a tenant-scoped settings system that lets each school (tenant) configure operational parameters through the UI — replacing
  hardcoded defaults and environment variables.

  **72 files changed**, ~7k lines added across backend, frontend, and documentation.

  ### What was done

  - **Planning & RFC** — settings system RFC, architecture review, and phased implementation plan
  - **Backend core** — type-safe settings registry, schema builder, service/repository layers, DB migration with RLS
  - **Backend integration** — scheduler, device auth, and checkout now read from tenant settings instead of env vars
  - **Frontend** — dynamic settings page rendered from backend schema, per-field save-on-blur with debounce auto-save, inline validation,
  green/red border feedback
  - **Profile extraction** — separated user profile (name, avatar, password) from workspace settings into its own `/profile` route
  - **Virtual device protection** — reserved web manual device can no longer be modified or deleted
  - **Comprehensive tests** — 180+ tests across backend and frontend (models, service, repositories, API handlers, field components, page
  rendering)

  ### How the settings system works

  #### Registry (compile-time)
  All settings are defined in `backend/services/config/defaults/` as typed Go structs registered in a global registry
  (`models/config/registry.go`). Each setting has:
  - A unique key (e.g. `operations.session_end_time`)
  - A type (boolean, number, time, text, select, password)
  - A default value, validation rules, category, and display metadata
  - Optional `DependsOn` for conditional visibility (e.g. time picker only shown when the parent toggle is enabled)

  Adding a new setting = one `Register()` call in a defaults file. No frontend code changes needed.

  #### Schema endpoint
  `GET /api/settings/schema` serializes the registry into a JSON schema grouped by tabs and categories (Operations, GDPR, Security). The
  frontend reads this once and dynamically renders all fields. Settings are filtered by the user's permissions — admin-only settings are hidden
  from non-admin users.

  #### Storage (per-tenant)
  `config.setting_values` stores tenant overrides (tenant_id + key + JSON value). If no override exists, the registry default is used. All
  changes are audit-logged to `config.setting_audit` with before/after values and the acting account.

  #### Tenant scoping
  The table has RLS policies. The service reads tenant ID from context (`tenant.FromContext(ctx)`) and all queries go through the tenant
  transaction middleware — same pattern as every other tenant-scoped table. Device auth uses `ResolveStringForTenant()` which wraps the query in
   its own `WithTenantTx` since device auth runs before tenant middleware.

  #### Frontend rendering
  `settings-page.tsx` fetches the schema, groups settings by category, and renders typed field components (boolean toggle, number input, time
  picker, select dropdown, password, text). Save behavior:
  - **Boolean/select**: save immediately on click
  - **Text/number/time**: save on blur or after 3 seconds of idle (debounce), with green border feedback for 4 seconds on success
  - **Password**: explicit save button (write-only, masked display)
  - **Optimistic updates**: values update locally without re-fetching the schema (no loading flicker), with a background sync after 6 seconds to
   catch server-side differences
  - **DependsOn**: evaluated client-side for instant show/hide when parent values change

  #### Integration points
  - **Scheduler** — refactored from global `os.Getenv()` to per-tenant minute-polling. `session_end_time`, `cleanup_time`, session cleanup
  interval, break auto-end interval now read from tenant settings at runtime via `resolveStringSetting`/`resolveBoolSetting`/`resolveIntSetting`
   with env var fallback
  - **Device auth** — OGS device PIN resolves per-tenant from `security.ogs_device_pin` setting with 30-second TTL cache, falling back to
  `OGS_DEVICE_PIN` env var
  - **Checkout helper** — daily checkout time reads from `operations.student_daily_checkout_time` setting with env var fallback

  ### ⚠️ Environment variable migration

  The following env vars are now configurable per-tenant via the settings UI. They still work as fallbacks during the transition, but should
  eventually be removed from `.env` / `docker-compose.yml`:

  | Setting key | Replaces env var | Default |
  |-------------|-----------------|---------|
  | `security.ogs_device_pin` | `OGS_DEVICE_PIN` | `""` |
  | `operations.session_end_enabled` | `SESSION_END_SCHEDULER_ENABLED` | `true` |
  | `operations.session_end_time` | `SESSION_END_TIME` | `18:00` |
  | `operations.session_end_timeout_minutes` | `SESSION_END_TIMEOUT_MINUTES` | `10` |
  | `operations.student_daily_checkout_time` | `STUDENT_DAILY_CHECKOUT_TIME` | `15:00` |
  | `operations.session_cleanup_enabled` | `SESSION_CLEANUP_ENABLED` | `true` |
  | `operations.session_cleanup_interval_minutes` | `SESSION_CLEANUP_INTERVAL_MINUTES` | `15` |
  | `operations.session_abandoned_threshold_minutes` | `SESSION_ABANDONED_THRESHOLD_MINUTES` | `60` |
  | `operations.break_auto_end_interval_seconds` | `BREAK_AUTO_END_INTERVAL_SECONDS` | `60` |
  | `gdpr.data_cleanup_enabled` | `CLEANUP_SCHEDULER_ENABLED` | `true` |
  | `gdpr.data_cleanup_time` | `CLEANUP_SCHEDULER_TIME` | `02:00` |
  | `gdpr.data_cleanup_timeout_minutes` | `CLEANUP_SCHEDULER_TIMEOUT_MINUTES` | `30` |

  **No env changes required for deployment** — all settings fall back to env vars when no tenant override exists. The env vars can be removed in
   a future cleanup PR after all tenants have been configured via the UI.

  ### Files of note

  | Area | Key files |
  |------|-----------|
  | Setting definitions | `backend/services/config/defaults/{operations,gdpr,security}.go` |
  | Registry & types | `backend/models/config/registry.go`, `setting_value.go`, `setting_audit.go` |
  | Service layer | `backend/services/config/settings_service.go` |
  | Schema builder | `backend/services/config/schema_builder.go` |
  | API handler | `backend/api/config/settings_api.go` |
  | Migration | `backend/database/migrations/001015024_config_setting_values.go` |
  | Scheduler refactor | `backend/services/scheduler/scheduler.go` (minute-polling) |
  | Device auth PIN | `backend/auth/device/device_auth.go` (PIN cache + RLS fix) |
  | Frontend page | `frontend/src/app/[tenant]/(protected)/settings/page.tsx` |
  | Frontend components | `frontend/src/components/settings/` |
  | API client | `frontend/src/lib/settings-api.ts` |
  | Next.js route handlers | `frontend/src/app/api/settings/{schema,values}/` |
  | Profile page (new) | `frontend/src/app/[tenant]/(protected)/profile/page.tsx` |

  ### Design documents

  | Document | Purpose |
  |----------|---------|
  | `docs/settings-system-rfc.md` | Original RFC (north star architecture) |
  | `docs/settings-v1-plan.md` | Trimmed v1 implementation plan with review fixes |
  | `docs/settings-rfc-review.md` | Yannick's review of the RFC |
  | `docs/settings-suggestions.md` | .env → settings migration mapping |
  | `docs/settings-implementation-plan.md` | Yannick's trimmed plan |
  | `docs/settings-v1-plan-review.md` | Yannick's review of the v1 plan |

### DIFF
--- a/backend/services/auth/invitation_service_test.go
+++ b/backend/services/auth/invitation_service_test.go
@@ -19,9 +19,9 @@ import (
 	baseModel "github.com/moto-nrw/project-phoenix/models/base"
 )
 
-// testStrongPassword is a valid password for unit tests that meets strength requirements.
+// testStrongCredential is a valid credential for unit tests that meets strength requirements.
 // This is NOT a real secret - it's only used with mocked services in tests.
-const testStrongPassword = "Str0ngP@ssword!" //nolint:gosec // Test-only constant, not a real credential
+const testStrongCredential = "Str0ngP@ssword!" //nolint:gosec // Test-only constant, not a real credential
 
 func newInvitationTestEnv(t *testing.T) (InvitationService, *stubInvitationTokenRepository, *stubAccountRepository, *stubRoleRepository, *stubAccountRoleRepository, *stubPersonRepository, *capturingMailer, sqlmock.Sqlmock, func()) {
 	service, invitations, accounts, roles, accountRoles, persons, _, mock, cleanup := newInvitationTestEnvWithMailer(t, nil)
@@ -333,8 +333,8 @@ func TestAcceptInvitationCreatesAccountAndPerson(t *testing.T) {
 	account, err := service.AcceptInvitation(ctx, "accept", UserRegistrationData{
 		FirstName:       "Katherine",
 		LastName:        "Johnson",
-		Password:        testStrongPassword,
-		ConfirmPassword: testStrongPassword,
+		Password:        testStrongCredential,
+		ConfirmPassword: testStrongCredential,
 	})
 	require.NoError(t, err)
 	require.NotNil(t, account)
@@ -370,8 +370,8 @@ func TestAcceptInvitationRollsBackOnError(t *testing.T) {
 	_, err := service.AcceptInvitation(ctx, "fail", UserRegistrationData{
 		FirstName:       "Jane",
 		LastName:        "Doe",
-		Password:        testStrongPassword,
-		ConfirmPassword: testStrongPassword,
+		Password:        testStrongCredential,
+		ConfirmPassword: testStrongCredential,
 	})
 	require.Error(t, err)
 	require.False(t, token.IsUsed(), "invitation should remain unused on failure")
@@ -533,8 +533,8 @@ func TestAcceptInvitationSecondAttemptFails(t *testing.T) {
 	account, err := service.AcceptInvitation(ctx, "second-attempt-token", UserRegistrationData{
 		FirstName:       "First",
 		LastName:        "User",
-		Password:        testStrongPassword,
-		ConfirmPassword: testStrongPassword,
+		Password:        testStrongCredential,
+		ConfirmPassword: testStrongCredential,
 	})
 	require.NoError(t, err)
 	require.NotNil(t, account)
@@ -544,8 +544,8 @@ func TestAcceptInvitationSecondAttemptFails(t *testing.T) {
 	_, err = service.AcceptInvitation(ctx, "second-attempt-token", UserRegistrationData{
 		FirstName:       "Second",
 		LastName:        "User",
-		Password:        testStrongPassword,
-		ConfirmPassword: testStrongPassword,
+		Password:        testStrongCredential,
+		ConfirmPassword: testStrongCredential,
 	})
 	require.Error(t, err)
 	require.True(t, errors.Is(err, ErrInvitationUsed), "Second acceptance should fail with ErrInvitationUsed")
@@ -636,8 +636,8 @@ func TestAcceptInvitationDeletedSchoolRejectsAcceptance(t *testing.T) {
 	_, err = service.AcceptInvitation(ctx, "deleted-school-token", UserRegistrationData{
 		FirstName:       "Ghost",
 		LastName:        "User",
-		Password:        testStrongPassword,
-		ConfirmPassword: testStrongPassword,
+		Password:        testStrongCredential,
+		ConfirmPassword: testStrongCredential,
 	})
 	require.Error(t, err)
 	require.True(t, errors.Is(err, ErrInvitationTenantDeleted),
@@ -671,8 +671,8 @@ func TestAcceptInvitationReusesExistingAccountForNewTenant(t *testing.T) {
 	account, err := service.AcceptInvitation(ctx, "existing-account", UserRegistrationData{
 		FirstName:       "Alex",
 		LastName:        "Principal",
-		Password:        testStrongPassword,
-		ConfirmPassword: testStrongPassword,
+		Password:        testStrongCredential,
+		ConfirmPassword: testStrongCredential,
 	})
 	require.NoError(t, err)
 	require.NotNil(t, account)


yungweng:
Hey @christiank119! Great work on this — the settings system is a solid foundation and the amount of ground covered here is impressive. I ran a thorough review and found a few issues that need to be addressed before we can merge. Nothing unfixable, but some of these are correctness/security regressions that would bite us in production.

---

## 🔴 P1 — Must fix before merge

### 1. Per-setting write permissions not enforced on update/reset
**`backend/api/config/settings_api.go:47-49`**

The `PUT` and `DELETE` endpoints only require the coarse `config:update` permission, but several settings (GDPR cleanup, `security.ogs_device_pin`) declare `WritePermission: "config:manage"`. Since `SetValue`/`ResetValue` never re-check the definition's `WritePermission`, any user with `config:update` can mutate admin-only settings via a direct API call — even though the schema correctly hides them in the UI.

**Fix:** Check `definition.WritePermission` in the handler (or service layer) before allowing writes, and return 403 if the caller lacks that permission.

### 2. No type validation for non-numeric settings
**`backend/services/config/settings_service.go:253-274`**

`validateValue()` only checks numeric ranges. Boolean, time, select, text, and password settings accept arbitrary JSON. This means:
- `"yes"` is accepted for `operations.session_end_enabled` (resolves as `false`, silently disabling the scheduler)
- `"25:99"` is accepted for `operations.session_end_time` (never matches, scheduler silently stops)
- Select fields accept values outside their allowed options

**Fix:** Add type-specific validation — booleans must be actual bools, time fields must match `HH:MM`, select fields must be one of the defined options, etc.

### 3. Device PIN setting not wired to device auth
**`backend/services/config/defaults/security.go:8-15`** + **`backend/auth/device/device_auth.go`**

`security.ogs_device_pin` is registered as a writable tenant setting, but device authentication still validates `X-Staff-PIN` against `os.Getenv("OGS_DEVICE_PIN")`. An admin can "rotate" the PIN in the settings UI without changing what `/api/iot/*` actually accepts. This is dangerous — the old PIN stays active and the new one appears broken.

**Fix:** Either wire `device_auth.go` to read from the settings service (with env var fallback), or remove this setting from the registry until it's actually consumed.

### 4. PIN must be exactly 4 digits
**`backend/services/config/defaults/security.go`**

The PIN field accepts arbitrary length input, but PyrePortal (the only consumer) hardcodes a 4-digit PIN pad. If an admin sets a 6-digit PIN via the settings UI, no device will ever be able to enter it.

**Fix:** Add validation that `security.ogs_device_pin` must be exactly 4 numeric digits (or empty).

### 5. Cleanup scheduler changes from opt-in to always-on
**`backend/services/scheduler/scheduler.go:248-250`**

Since `services.Factory` now always provides `SettingsService`, the early-return guard (`if s.settingsService == nil`) is never reached. The registry default for `gdpr.data_cleanup_enabled` is `true`, so deployments that previously left `CLEANUP_SCHEDULER_ENABLED` unset (= disabled) will start running the daily cleanup/deletion job immediately after this patch.

**Fix:** The env var fallback logic needs to check whether the env var was explicitly set and prefer that value. Or ensure the default matches the previous behavior (disabled unless explicitly enabled).

### 6. Existing env var overrides silently ignored
**`backend/services/scheduler/scheduler.go:931-939`**

The `resolve*Setting` helpers consult the settings service first. Because `Resolve*` returns the registry default even when no tenant override exists, existing deployments that rely on non-default env var values (e.g. `SESSION_END_TIME=16:30`) will silently fall back to the hardcoded defaults (`18:00`) after this patch — unless every tenant manually creates a DB override first.

**Fix:** The fallback chain should be: tenant DB override → env var → registry default. Currently it's: tenant DB override → registry default (env var skipped).

---

## 🟡 P2 — Should fix

### 7. Break auto-end interval hardcoded to 60s
**`backend/services/scheduler/scheduler.go:878`**

The polling ticker is hardcoded to `60 * time.Second`, ignoring the `operations.break_auto_end_interval_seconds` setting entirely. Changing this setting in the UI has no effect.

### 8. Daily checkout time setting not wired to IoT check-in
**`backend/services/config/defaults/operations.go:64-75`** + **`backend/api/iot/checkin/helpers.go`**

`operations.student_daily_checkout_time` is exposed as a tenant-editable setting, but the IoT check-in flow still reads `os.Getenv("STUDENT_DAILY_CHECKOUT_TIME")`. Same problem as the PIN — changing it in the UI does nothing.

### 9. Empty password default masked as configured
**`backend/services/config/schema_builder.go:45-47`**

Fresh tenants resolve `security.ogs_device_pin` to the registry default `""`, but the schema builder masks every non-nil password value (including empty string) as `"••••••"`. Admins can't tell whether a PIN is actually configured on a new tenant.

**Fix:** Only mask non-empty password values.

### 10. Frontend loading race — settings page can stay permanently empty
**`frontend/src/components/settings/settings-page.tsx:202-206`**

The fetch effect fires once on mount, but `fetchSettingsSchema()` returns `null` when the auth token isn't ready yet. The effect never re-runs after `useSession()` finishes loading, so an authenticated user can get stuck on "Keine Einstellungen verfügbar." until they manually reload.

**Fix:** Add the session/token state to the effect's dependency array so it re-fetches once auth is ready.

---

## Summary

The core architecture is sound — the registry pattern, schema builder, and tenant-scoped storage are well designed. The main issues are around the integration seams: settings that aren't actually wired to the code they claim to control, missing type validation, and the env var → settings migration not preserving existing behavior.

I'd suggest tackling it in this order:
1. Fix the authorization gap (P1 #1)
2. Add type validation (P1 #2)
3. Fix the env var fallback chain (P1 #5, #6) — this is the most important for safe deployment
4. Either wire or remove the unwired settings (P1 #3, P2 #7, #8)
5. PIN length validation (P1 #4)
6. Frontend fixes (P2 #9, #10)

Happy to pair on any of these if you want to talk through the approach! 🙌

PR #1228, opened 2026-04-09
feat: custom login image upload for tenant settings

theitger:
## Summary
- Tenants can upload a custom image for their login page via a new **Personalization** tab in settings
- Login page conditionally renders the tenant's custom image or the default MOTO logo
- Public image-serving route with path traversal protection for unauthenticated access

### Why is this PR large?
- **Refactor (~700 lines moved):** Image upload/serve/validate logic was duplicated across `usercontext/api.go` and `staff/api.go`. Extracted into a shared `common/upload` module to reuse for login images without tripling the duplication.
- **Tests (62% of additions):** Integration and unit tests for the new endpoints, the shared upload module, and the frontend components.

## Test plan
- [ ] Tenant settings → Personalization tab → upload image → preview appears
- [ ] Delete the image → removed from settings and login page
- [ ] Login page shows custom image when set, default logo when not
- [ ] Existing avatar upload (staff, students) still works after refactor

### DIFF
--- a/backend/api/base.go
+++ b/backend/api/base.go
@@ -19,6 +19,7 @@ import (
 	activitiesAPI "github.com/moto-nrw/project-phoenix/api/activities"
 	adminAPI "github.com/moto-nrw/project-phoenix/api/admin"
 	authAPI "github.com/moto-nrw/project-phoenix/api/auth"
+	apiCommon "github.com/moto-nrw/project-phoenix/api/common"
 	configAPI "github.com/moto-nrw/project-phoenix/api/config"
 	databaseAPI "github.com/moto-nrw/project-phoenix/api/database"
 	feedbackAPI "github.com/moto-nrw/project-phoenix/api/feedback"
@@ -386,6 +387,12 @@ func (a *API) registerRoutesWithRateLimiting() {
 
 	// Note: Avatar files are served through authenticated endpoints, not as static files
 	// This prevents unauthorized access to user avatars
+
+	// Public login image serving (no auth — displayed on the login page before authentication)
+	a.Router.Get("/public/login-image/{filename}", func(w http.ResponseWriter, r *http.Request) {
+		filename := chi.URLParam(r, "filename")
+		apiCommon.ServeImage(w, r, "public/uploads/login-images", filename, "public, max-age=86400")
+	})
 
 	// Mount API resources
 	// Auth routes mounted at root level to match frontend expectations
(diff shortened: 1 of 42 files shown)


yungweng (commented):
Hey @theitger — ran this through three independent reviews (mine + two automated passes) and aggregated the findings. Overall this is a really solid PR, especially the `api/common/upload` extraction which cleans up real duplication without leaving dead code behind, and the test coverage at 62% of additions is well above average. Security posture is strong with defense-in-depth on path traversal and magic-byte validation. No must-fix issues found across any of the reviews, which is great.

That said, there are a few should-fix items worth addressing before merge:

---

**1. Frontend rejects valid `.jfif` JPEG uploads**
`frontend/src/app/api/settings/login-image/route.ts` — the `allowedExtensions` whitelist is `[".jpg", ".jpeg", ".png", ".webp"]`, but Windows and Chrome commonly save JPEGs with a `.jfif` extension. The backend's `ParseImage()` would happily accept these by magic bytes and store them as `.jpg`, but the frontend blocks them with a 415 before the request ever reaches the backend — so this is a regression in the new feature that'll hit some admins on Windows machines. Adding `.jfif` to the whitelist should do the trick.

**2. `getLoginImage` returns 200 with `canEdit=true` when there's no tenant context**
`backend/api/config/settings_api.go` around line 148–154 — when `tenant.FromContext()` returns 0 (e.g. an org-scoped session), `GetLoginImageURL(0)` returns `""` which is treated as success, and the handler computes `canEdit` purely from permissions. The Personalization tab then shows upload/delete buttons that will always fail with 400 "no tenant context" on POST/DELETE — so the GET endpoint is advertising capabilities the user can never actually use. Should fail fast with 400 here too, matching the guards on the write endpoints.

**3. Missing upload field returns 500 instead of 4xx**
`frontend/src/app/api/settings/login-image/route.ts` around line 80–82 — if the client submits multipart form data without the `login_image` field, `throw new Error("No image file provided")` falls through to the generic 500 handler. That makes a simple client mistake look like a server error and will pollute error logs over time. Returning a 400 or 422 here keeps the contract consistent with how the backend treats the same condition.

**4. Orphan files on crash between file save and DB commit (minor — document or defer)**
In `uploadLoginImage` the file is saved to disk first, then the DB is updated in a separate `WithAdminTx`. If the process crashes between the two — or two concurrent uploads race and one loses — the loser's file becomes an orphan on disk. For login images this isn't catastrophic, but it's worth either adding a comment documenting the known limitation or tracking a follow-up for a periodic cleanup job.

---

None of these are blockers on their own, but items 1–3 are small surgical fixes that'd be nice to have before merge. Great work on the refactoring and the test coverage — the hermetic patterns are followed correctly and the `FileValidationError` class with semantic HTTP status codes is a nice touch.

Issue #1187, opened 2026-04-01
Announcements targeting system roles don't reach users with custom tenant roles


theitger:
## Summary

Operator announcements targeted at system roles (`admin`, `user`, `guardian`) are invisible to users who only have custom tenant-created roles. The announcement filtering uses strict string matching on role names, with no concept of role equivalence or hierarchy.

## Problem

The announcement query in `announcement_view_repository.go` filters by PostgreSQL array overlap:

```sql
AND (a.target_roles = '{}' OR a.target_roles && ?::text[])
```

The operator UI only offers the three hardcoded system roles as targeting options (`admin`, `user`, `guardian`). Custom tenant roles (e.g. "Pädagoge", "Gruppenleitung") are never included in this check.

**Example:**
- Operator publishes announcement targeting Betreuer (`target_roles: ['user']`)
- Staff member with only a custom role `"Gruppenleitung"` → JWT `roles: ["Gruppenleitung"]`
- `['user'] && ['Gruppenleitung']` → no overlap → **announcement is never shown**

## Expected Behavior

Announcements targeting a system role should also reach users whose custom tenant roles are functionally equivalent (e.g. a custom role created as a Betreuer-level role should still receive Betreuer-targeted announcements).

## Possible Approaches

1. **Map custom roles to a base system role** — add a `base_role` or `parent_role` field to custom roles so the system knows "Gruppenleitung" is a variant of "user". Query expands to check both.
2. **Include custom roles in targeting UI** — let operators also select tenant-specific custom roles when creating announcements. More flexible but more complex UI.
3. **Require system role retention** — enforce that every user always keeps their system role, making custom roles purely additive. Simplest fix but limits tenant flexibility.

## Affected Files

- `backend/database/repositories/platform/announcement_view_repository.go` — filtering query
- `backend/models/platform/announcement.go` — hardcoded role constants
- `frontend/src/lib/operator/announcements-helpers.ts` — `SYSTEM_ROLE_LABELS` (only 3 roles)
- `backend/api/operator/announcements.go` — operator creation endpoint

theitger:
Resolved via PR #1233 — custom tenant roles now map to system roles via `base_role`, so announcements targeting system roles reach users with equivalent custom roles. Closing.

PR #1233, opened 2026-04-09
fix: map custom tenant roles to system roles for announcement targeting (#1187)

theitger:
## Summary

- Adds a `base_role` column to `auth.roles` so every custom tenant role **must be** mapped to a system role (`admin`, `user`, `guardian`) for announcement delivery
- Updates the announcement unread query and stats query to match users via `base_role` when their JWT role name doesn't directly match `target_roles`
- Exposes `base_role` in the role CRUD API — required on create, optional on update (preserves existing value when omitted), system roles are guarded from receiving a `base_role`
- Frontend: role create/edit form includes a required "Systemrolle" select field; role list and detail modal display the mapped base role

## Warning for existing custom roles

Pre-existing custom roles (created before this migration) will have `base_role = NULL`. The roles page shows a prominent amber warning banner:

> **"{n} Rollen haben keine Systemrollen-Zuordnung"**
> Ankündigungen werden möglicherweise nicht korrekt zugestellt. Bitte bearbeiten Sie die betroffenen Rollen und wählen Sie eine Systemrolle aus.

Individual unclassified roles are tagged with an amber **"Zuordnung fehlt"** badge in the role list.

## Migration

`001015031_roles_base_role` — adds `base_role VARCHAR` column with a `CHECK` constraint (`admin`, `user`, `guardian`) and a partial index on non-null values. Existing rows are left `NULL` intentionally (requires manual classification per tenant).

## Test plan

- [ ] Run migration on fresh and existing databases
- [ ] Verify announcement delivery reaches users with custom roles mapped via `base_role`
- [ ] Verify the warning banner appears when unclassified custom roles exist and disappears once all are mapped
- [ ] Verify system roles cannot be assigned a `base_role` via API
- [ ] Run `go test ./...` — hermetic test suite passes

### DIFF
--- a/backend/api/auth/api.go
+++ b/backend/api/auth/api.go
@@ -907,8 +907,8 @@ func normalizeBaseRole(br *string) *string {
 
 // validateBaseRoleValue checks that a non-nil base_role value is one of the allowed system roles.
 func validateBaseRoleValue(value string) error {
-	if !slices.Contains(authModel.ValidBaseRoles, value) {
-		return fmt.Errorf("base_role must be one of: %v", authModel.ValidBaseRoles)
+	if !slices.Contains(authModel.ValidBaseRoles(), value) {
+		return fmt.Errorf("base_role must be one of: %v", authModel.ValidBaseRoles())
 	}
 	return nil
 }
@@ -917,7 +917,7 @@ func validateBaseRoleValue(value string) error {
 // This runs at the request layer to return 400 (not 500) for invalid payloads.
 func validateBaseRole(br *string) error {
 	if br == nil {
-		return fmt.Errorf("base_role is required; must be one of: %v", authModel.ValidBaseRoles)
+		return fmt.Errorf("base_role is required; must be one of: %v", authModel.ValidBaseRoles())
 	}
 	return validateBaseRoleValue(*br)
 }
--- a/backend/api/auth/api.go
+++ b/backend/api/auth/api.go
@@ -4,9 +4,11 @@ import (
 	"database/sql"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"log/slog"
 	"net"
 	"net/http"
+	"slices"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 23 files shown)


yungweng:
Hey @theitger, nice work on this one! The base_role mapping approach is solid and the test coverage is thorough. A few things came up during review:

## Must Fix

**Migration dependency `1.15.30` does not exist** — `001015031_roles_base_role.go:20`

`DependsOn` references `"1.15.30"` but the highest migration in the tree is `1.15.29`. `ValidateMigrations()` will reject this, which means `./main migrate` fails on every environment — fresh setup, staging, production. This blocks deployment. Should be `"1.15.29"` (or whichever migration you actually depend on).

## Should Fix

1. **Cannot clear `base_role` back to NULL via update API** — `api.go:1176-1180`
   `normalizeBaseRole` converts `""` → `nil`, and the handler treats `nil` as "preserve existing". Once set, there's no way to unset it through the API. If that's intentional, worth a comment. If not, needs a way to distinguish "not sent" from "clear this".

2. **Service layer doesn't enforce required `base_role` on create** — `role_management.go:16`
   Comment says "baseRole is required" but the service happily accepts `nil`. Only the HTTP handler validates. Any future internal caller (provisioning, seeder) could create roles without base_role, silently breaking announcement targeting. A one-liner validation in the service would close this gap.

3. **No deduplication test for `GetStats()`**
   `GetUnreadForUser` has a great test proving dual-match users aren't double-counted. `GetStats()` has the same dual-match logic but no equivalent test. `COUNT(DISTINCT)` handles it correctly today — a test would prevent regressions.

## Looks Good

- SQL is correct and properly tenant-scoped (`ar.tenant_id = ?` in the EXISTS subquery)
- The cross-layer sync test (`base_role_sync_test.go`) is a smart way to keep Go, DB constraint, and frontend labels in lockstep
- Migration is safe (nullable column, no table rewrite, idempotent DDL)
- Frontend warning banner for unclassified roles is great UX
- Test modifications in `auth_core_test.go` / `refactored_modules_test.go` are mechanical signature updates (not assertion changes) — fine by me

theitger:
Thanks for the review @yungweng!

**Points 1 & 3** — fixed in `171e560d`: migration dependency corrected to `"1.15.29"`, and `CreateRole` now validates `base_role` at the service layer too.

**Point 2 (cannot clear base_role)** — This is intentional. Once a base_role is set, clearing it would put the role back into the "unclassified" state and break announcement targeting for users with that role. I added a comment in `normalizeBaseRole` explaining the intent. That said, if you'd prefer it to be clearable, I can add that — let me know.

**Point 4 (GetStats deduplication test)** — This test already exists: `announcement_view_repository_test.go:1315` — `"account with both direct role and base_role-mapped role counted once"`. It creates an account with both a direct "user" role and a custom role with `base_role = "user"`, then asserts `stats.TargetCount == 1`.

yungweng (approved):
mal gucken was passiert

PR #1234, opened 2026-04-10
feat: protect system rooms and activities from deletion/renaming

fl0r14n28:
## Summary
- Schulhof/WC rooms and Schulhof Freispiel/WC activities are system entities required by PyrePortal (kiosk app)
- Adds backend guards (HTTP 403) preventing deletion and renaming of these entities
- Adds frontend guards: delete button hidden, name field disabled in edit modals
- Backend is the authoritative guard — frontend is UX-only

## Changes

**Backend (7 files)**
- `constants/activities.go` — `IsSystemRoomName()`, `IsSystemActivityName()` helpers
- `services/facilities/` — `ErrSystemRoomProtected` + guards in `DeleteRoom()` and `UpdateRoom()`
- `services/activities/` — `ErrSystemActivityProtected` + guards in `DeleteGroup()` and `UpdateGroup()`
- `api/rooms/errors.go`, `api/activities/errors.go` — map new errors to HTTP 403

**Frontend (9 files)**
- `room-helpers.ts`, `activity-helpers.ts` — `isSystemRoom()`, `isSystemActivity()` helpers
- `detail-modal-actions.tsx` — new `hideDelete` prop
- Room/Activity detail + edit modals — hide delete, disable name field for system entities
- `database-form.tsx`, `types.ts` — `disabled` prop on `FormField`

## Test plan
- [x] Verify Schulhof/WC rooms cannot be deleted via UI (button hidden)
- [x] Verify Schulhof/WC rooms cannot be renamed via UI (field disabled)
- [x] Verify Schulhof Freispiel/WC activities cannot be deleted via UI (button hidden)
- [x] Verify Schulhof Freispiel/WC activities cannot be renamed via UI (field disabled)
- [x] Verify backend returns 403 when attempting delete/rename via API directly
- [x] Verify other room/activity properties (capacity, building, etc.) remain editable
- [x] Verify non-system rooms/activities are unaffected

### DIFF
--- a/backend/api/activities/errors.go
+++ b/backend/api/activities/errors.go
@@ -50,6 +50,8 @@ func ErrorRenderer(err error) render.Renderer {
 			return ErrorConflictWithCode(actErr, errorCodeOnlySupervisorReplacementRequired)
 		case errors.Is(actErr, activities.ErrNotEnrolled):
 			return ErrorNotFound(actErr)
+		case errors.Is(actErr, activities.ErrSystemActivityProtected):
+			return ErrorForbidden(actErr)
 		case errors.Is(actErr, activities.ErrGroupClosed):
 			return ErrorForbidden(actErr)
 		case errors.Is(actErr, activities.ErrInvalidAttendanceStatus):
--- a/backend/api/activities/errors_test.go
+++ b/backend/api/activities/errors_test.go
@@ -63,12 +63,24 @@ func TestErrorRenderer_ConflictErrors(t *testing.T) {
 }
 
 func TestErrorRenderer_ForbiddenErrors(t *testing.T) {
-	actErr := &activities.ActivityError{Err: activities.ErrGroupClosed}
-	renderer := activitiesAPI.ErrorRenderer(actErr)
-	resp, ok := renderer.(*activitiesAPI.ErrorResponse)
-	assert.True(t, ok)
-	assert.Equal(t, http.StatusForbidden, resp.HTTPStatusCode)
-	assert.Equal(t, "error", resp.Status)
+	tests := []struct {
+		name    string
+		baseErr error
+	}{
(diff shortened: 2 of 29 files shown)


yungweng:
@fl0r14n28 bitte erst review requesten sobald gesamte CI fertig. Weil sonst Spam :(

yungweng (commented):
Hey @fl0r14n28, nice work on this! The layering (backend authoritative + frontend UX) is spot on and the test coverage is solid. Two things worth addressing:

**1. Extra DB query in `UpdateGroup` (activity_service.go:260)**
`FindByID` runs on every update now, but `CanModifyActivity` already fetches the group internally — so we're hitting the DB twice. Could we restructure so `CanModifyActivity` returns the group it found, or move the system name check after that call?

**2. Frontend system name lists duplicated from backend constants**
`SYSTEM_ROOM_NAMES` and `SYSTEM_ACTIVITY_NAMES` are hardcoded in the frontend helpers with a "must stay in sync" comment. If we ever add a third system entity on the backend, the frontend won't know. Worth a follow-up to either expose an `is_system` flag in the room/activity API response or add a shared endpoint — that way the backend stays the single source of truth.

Everything else looks good to me!

fl0r14n28:
Thanks for the review!

**1. Extra DB query:** Moved the system name check after `CanModifyActivity` so admins (who short-circuit without a DB hit) don't pay the cost. The rename guard still does one `FindByID` — changing `CanModifyActivity`'s return signature would touch 5+ callers, not worth the churn for a PK lookup on updates only. → b7796382

**2. Frontend name lists:** Agreed. Follow-up: expose an `is_system` flag in the room/activity API response so the backend stays the single source of truth. Will create an issue for that.

yungweng (approved):
lgtm

PR #1241, opened 2026-04-11
feat: Admin supervision overview (opt-in per tenant)

fl0r14n28:
## Summary

- New per-tenant setting `operations.admin_supervision_overview` (default: off) that lets admins see all active supervisions
- When enabled, admins see all active rooms in the sidebar accordion, can view students, and interact (check-in/out)
- SSE live updates for all active groups when the setting is active
- No impact on existing caregiver (user/teacher) workflows

### Backend
- New setting registered in operations/sessions category
- `GET /api/active/supervisors/all` endpoint with admin + setting validation
- `FindAllActive` repository method for all active supervisions
- SSE admin subscription to all active groups when setting enabled
- Admin bypass for visit display authorization

### Frontend
- `SupervisionContext` fetches all groups for admins via new endpoint
- Sidebar accordion visible for admins with active rooms
- Custom `ActiveSupervisionGate` replaces `RoleGuard` on the page
- Admin fallback in BFF dashboard endpoint
- Mobile nav adds "Aufsicht" tab dynamically for admins

### Simulator fix
- Each activity now assigned to a unique room (matches production constraint)
- Room ID passed explicitly in session start requests

## Test plan

- [x] Login as admin, enable setting in Settings > Betrieb > "Administrator-Aufsichtsübersicht"
- [x] Verify all active rooms appear in sidebar accordion
- [x] Click a room — students are displayed with check-in times
- [x] Disable setting — accordion disappears, page shows "Kein Zugriff"
- [x] Login as caregiver (user/teacher role) — behavior unchanged
- [x] Verify SSE live updates work for admin (check in a student via another session)
- [x] Verify tenant isolation — admin only sees their own school's supervisions


yungweng (commented):
@fl0r14n28 Great work on this feature — the architecture is clean, test coverage is solid (94.5%), and the security bypass fix in the second commit shows good diligence. Here are the findings from review, grouped by priority.

---

## P1 — Schulhof bypasses the feature flag

**`frontend/src/app/[tenant]/(protected)/active-supervisions/page.tsx` — `ActiveSupervisionGate`**

The gate lets admins through when `supervisedRooms.length > 0`. But the Schulhof integration always injects a synthetic room entry into `supervisedRooms` whenever a Schulhof exists in the tenant — regardless of the `admin_supervision_overview` setting.

This means an admin in any school with a Schulhof can open `/active-supervisions` even when the setting is OFF. The feature flag gets bypassed entirely.

**Suggestion:** Distinguish "rooms from admin overview" from "synthetic Schulhof entry" in the gate, or check the setting value directly.

---

## P1 — Admin dashboard loads once, then never updates

**`backend/api/sse/sse_helpers.go` + `supervisors_handlers.go` + frontend SSE wiring**

Three things conspire here:

1. The HTTP endpoint (`/supervisors/all`) returns data from `ListActiveGroups` — includes unclaimed rooms (no supervisor).
2. SSE uses `GetAllActiveSupervisions` — queries `active.group_supervisors`, so unclaimed rooms are excluded from subscriptions.
3. For admin-only accounts (no staff record), `resolveStaff()` may fail before SSE topics are built, and `useGlobalSSE()` may refuse to connect without the `user` role.

**Result:** Admins see all rooms on page load, but student check-in/check-out events never arrive for rooms they don't personally supervise. Counts stay frozen until manual refresh.

**Suggestion:** Align the HTTP and SSE paths to use the same data source, or subscribe admins to all active group IDs from `active.groups`.

---

## P2 — Dual-role users (admin + teacher) see all rooms

**`frontend/src/lib/supervision-context.tsx:215-223`**

The supervision context sends all admins to `/api/active/supervisors/all`. But a teacher who also has admin access currently sees only their own rooms. After this PR with the setting enabled, they suddenly see every room in the sidebar.

The rest of the PR correctly distinguishes dual-role users — mobile nav checks `!isCaregiver(session)` before injecting the admin tab — but the data fetch doesn't apply the same condition.

**Suggestion:** Gate the admin endpoint fetch behind `isAdmin && !isCaregiver`, matching the condition used elsewhere.

---

## P2 — Inconsistent error handling between supervision context and BFF

**`frontend/src/lib/supervision-context.tsx:226-232` vs `route.ts:177-188`**

The supervision context only falls back to the staff endpoint on `403`. A `500` or timeout means the admin loses their own rooms too (regression).

The BFF route has the opposite problem — it `.catch()`es everything, so a real backend outage silently returns an empty dashboard instead of surfacing the error.

**Suggestion:** Fall back on `!adminResponse.ok` in the supervision context. In the BFF, limit the catch-and-fallback to `403` so real errors propagate.

---

## P3 — Unnecessary 403 for every caregiver dashboard load

**`frontend/src/app/api/active-supervision-dashboard/route.ts:177-185`**

The BFF now tries `/api/active/supervisors/all` first for every request, including normal caregivers. For non-admins this always fails with `403`, adding an extra backend round-trip on the hot path and creating noisy log entries.

**Suggestion:** Check the session roles before choosing which endpoint to call.

---

## Minor observations

- **Redundant `IsActive()` filtering** — the repo query, service, and handler all filter for active records (3 layers). Works, but a comment explaining the defense-in-depth intent would help.
- **Endpoint naming** — `/supervisors/all` returns active groups, not supervisors. Slightly misleading.
- **~200-line mock stubs duplicated** across `supervision_overview_internal_test.go` and `resolve_supervisions_internal_test.go`. Worth considering a shared helper as the interface grows.
- **No positive-path test** proving an admin with the setting enabled can view visits for a group they don't supervise. Only the forbidden paths are tested.
- **Route ordering** — `/all` registered after `/{id}` in Chi. Works correctly (literals beat params), but placing it before `/{id}` would make the intent clearer.

---

Overall this is a well-structured feature. The P1s (Schulhof bypass + stale SSE) are the ones I'd want addressed before merge. Happy to discuss any of these — nice work getting this far! 🚀

yungweng (approved):
Jep klappt alles!

PR #1285, opened 2026-04-17
feat: organization soft-delete with hierarchy guards

theitger:
## Summary

- Adds soft-delete + restore for organizations (Träger), mirroring the existing school pattern
- Org soft-delete blocked while any non-deleted child school exists (409 with school count in the message)
- Schools cannot be created under, moved to, or restored under a soft-deleted org
- `UpdateOrganization` rejected on a soft-deleted org
- Announcement targeting cannot reference soft-deleted orgs/schools on create or when a new ID is added on update; historical IDs already on an announcement round-trip verbatim so unrelated edits don't silently widen scope to global

## New API Endpoints

| Method | Path | Purpose |
|--------|------|---------|
| DELETE | `/operator/organizations/{id}` | Soft-delete organization |
| POST | `/operator/organizations/{id}/restore` | Restore organization |

## New Error Types → HTTP Status

| Error | Status | When |
|-------|--------|------|
| `OrganizationAlreadyDeletedError` | 409 | Soft-delete on already-deleted org |
| `OrganizationNotDeletedError` | 409 | Restore on non-deleted org |
| `OrganizationHasSchoolsError` | 409 | Delete blocked — child schools still live |
| `OrganizationDeletedError` | 409 | School create / move / restore targets a deleted org; or `UpdateOrganization` on a deleted org |

### Backend
- Migration `1.15.32`: `deleted_at TIMESTAMPTZ` on `platform.organizations` + `idx_organizations_not_deleted` and `idx_schools_org_not_deleted` (both partial on `deleted_at IS NULL`)
- `Organization` model gains `DeletedAt` (JSON `omitempty`) and `IsDeleted()`
- `OrganizationRepository`: new `SoftDelete`, `Restore`, `FindByIDForShare`, `FindByIDForUpdate`; `CountByIDs` now filters `deleted_at IS NULL`
- `SchoolRepository`: new `CountNonDeletedByOrganizationID`; `CountByIDs` now filters `deleted_at IS NULL`; both repos share a private `findByIDWithLock(ctx, id, lockClause, op)` helper
- `OperatorProvisioningService`: new `SoftDeleteOrganization` / `RestoreOrganization`; `validateSchoolCreate`, `UpdateSchool` (org change), `RestoreSchool`, `UpdateOrganization` take the parent org `FOR SHARE` and reject on `IsDeleted()`; rows-affected-mismatch on the org UPDATE is remapped to the already/not-deleted error
- `AnnouncementService.UpdateAnnouncement` validates only newly-added target IDs via `validateNewTargetingIDs` (diff against existing record); create-path still validates the full set

### Frontend
- `soft-delete-shared.tsx`: `useSoftDeletable<T>` hook + shared `DeletedEntityCard`, `SoftDeleteConfirmationModal`, `RestoreConfirmationModal`, parametrized on German article/pronoun/possessive
- Provisioning page has a per-section Papierkorb toggle for orgs and schools; soft-delete modal disables confirm when the org still has live schools; restore modal disables confirm when the parent org is deleted (with visible reason)
- Deleted-school card reads `school.organization?.deletedAt || deletedOrgIds.has(school.organizationId)` so the restore button flips live on parent-org delete/restore without waiting for a schools refetch
- Announcements page filters active orgs/schools out of the pickers but preserves historical target IDs in form state; warning banner surfaces any selected target whose org or school is now soft-deleted
- New proxy routes: `DELETE /api/operator/provisioning/organizations/[id]`, `POST .../restore`
- New API wrappers: `softDeleteOrganization(id)`, `restoreOrganization(id)` with `encodeURIComponent`
- `revalidateTenantCache(slugs)` helper consolidates the inlined fire-and-forget revalidate-tenant fetches on the provisioning page and school modals

## Test plan

- [x] `go test ./backend/database/repositories/platform/organization_locking_test.go` — real-DB `SET LOCAL lock_timeout`, asserts FOR UPDATE↔FOR SHARE matrix + FOR SHARE parallelism
- [x] Service-layer tests cover every branch of `SoftDeleteOrganization` / `RestoreOrganization` (already/not deleted, has-schools, count error, rows-affected-mismatch, parent-org lookup failure on school restore)
- [x] Announcement service: historical deleted-org/school target round-trip on unrelated update; rejection of newly-added deleted-org target
- [x] Org/school repo tests pin `deleted_at IS NULL` on `CountByIDs` and `CountNonDeletedByOrganizationID`
- [x] Frontend route tests for `DELETE` + `restore` (happy, 401, invalid id, 409) and API-wrapper URL-encoding tests
- [ ] Manual: delete org with live schools → 409 shows count; delete empty org → Papierkorb → restore brings it back
- [ ] Manual: attempt school restore whose parent is deleted → confirm button disabled with reason
- [ ] Manual: edit announcement with a target that points at a now-deleted org → warning banner shows, unrelated save succeeds without widening scope


yungweng:
@theitger funktioniert das alles?

yungweng (approved):
## Mental model

> Soft-delete is a reversible tombstone. Locks enforce the "no orphans" invariant at write time. Validation filters `deleted_at` at read/check time. Historical data round-trips — new writes must point at live targets.

Every piece of the PR falls out of those three sentences.

## Review notes

- **Locking contract** is the strongest part. FOR UPDATE on parent org during soft-delete ↔ FOR SHARE on parent during any school mutation, with the pre-check (`CountNonDeletedByOrganizationID`) running *inside* the locked tx. Lock order org → school, no deadlock. Real-DB `lock_timeout` test makes regressions deterministic instead of timing-dependent.
- **Announcement diff** (`validateNewTargetingIDs`) correctly preserves historical scope: create-path validates all, update-path validates only newly-added IDs. Unrelated edits on an announcement whose targets now point at deleted orgs do not silently widen scope to global.
- **Rows-affected-mismatch remap** to `AlreadyDeleted` / `NotDeleted` on SoftDelete/Restore handles the concurrent-click race cleanly.
- **Partial indexes** on `deleted_at IS NULL` are correct for the hot paths.
- All new endpoints under `RequiresOperatorScope`, audit-logged with operator ID + IP.

## Non-blocking nits

- PR description references migration `1.15.32` but the last commit bumped it to `1.15.35` — worth updating the body before merge.
- Description says "rows-affected-mismatch on the org UPDATE is remapped to the already/not-deleted error"; in `UpdateOrganization` it maps to `InvalidDataError`. The remap only applies to SoftDelete/Restore. Practically unreachable thanks to FOR SHARE, but the description overclaims.
- `FindBySlug` / `FindByOrganizationAndSlug` don't filter `deleted_at IS NULL`. Keeping slugs reserved during soft-delete is arguably correct for safe restore — worth a one-line comment calling out the intent, like `FindBySubdomain` already does.
- `soft-delete-shared.tsx` uses generic Tailwind reds (`bg-red-50/50`, `border-red-100/50`, `bg-red-100`, `text-red-700`). Matches 31 existing occurrences across operator pages, so no new regression — but a consolidated shared component is the right place to fix.

LGTM.

PR #1332, opened 2026-04-28
fix(test): cover three more FK chains in cleanup (follow-up to #1314)

creintjes:
Follow-up to #1314, which was merged before review feedback could land.

#1314 fixed the silent-no-op cleanup pattern, but **three FK chains were still incomplete**: deleting a staff/room/device tripped a child-table FK silently and left orphaned parent rows behind. Each consecutive run accumulated one more orphan, surfacing later as flaky `expected N, got N+1` assertions.

Yannick verified locally with \`-count=2\` (two consecutive runs against the same DB) and saw the linear leak pattern: Run 1 → 1 leak, Run 2 → 2 leaks, Run 3 → 3 leaks. This PR closes the chains he identified.

## Summary

- **\`activities.groups.created_by → users.staff\`**: cleanup now also matches \`OR created_by = ?\` so deleting a staff fixture removes the activity groups it created.
- **\`active.groups.room_id → facilities.rooms\`**: cleanup now also matches \`OR room_id = ?\` so deleting a room removes the active groups bound to it.
- **\`active.attendance.device_id → iot.devices\`**: cleanup now also matches \`OR device_id = ?\` so deleting a device removes its attendance rows.
- **Fail-fast on system-staff fixture**: the \`users.persons (id=1)\` and \`users.staff (id=1)\` inserts in \`SetupTestDB\` previously used \`_, _ = db.ExecContext(...)\`, swallowing any error silently. Replaced with \`require.NoError(t, err, ...)\` so schema drift surfaces at its source instead of as cryptic 'missing staff' errors hundreds of layers down (per CLAUDE.md no-silent-fallbacks rule).

## Verification

```bash
docker compose --profile test down postgres-test -v
docker compose --profile test up -d postgres-test
APP_ENV=test docker compose run server ./main migrate
go test -p 1 -count=2 ./...   # all packages green, EXIT=0
```

\`-count=2\` runs the entire suite twice in the same DB without a reset. Both passes are now fully green — no leaks, no flakes.

## Out of scope (intentionally not touched)

- The \`_, _ =\` calls on \`facilities.rooms (id=1)\` and the sequence-advance block in \`SetupTestDB\` use the same silent-error pattern. Same reasoning as the staff fix would apply, but adding them here is scope creep. Worth a separate small cleanup if anyone agrees.
- The block-comment bypass in the guard test regex (\`hermetic_verification_test.go\`) — \`grep -rn "/\*" backend/test/*.go\` returns zero matches, so the existing self-skip is sufficient. Happy to harden if a future test file makes it relevant.

## Why this is a separate PR

#1314 was merged accidentally before this round of review feedback could be addressed. Rather than force-pushing or reverting, this stands on its own as a follow-up.

Refs #1296

### DIFF
--- a/backend/test/fixtures.go
+++ b/backend/test/fixtures.go
@@ -409,10 +409,12 @@ func CleanupActivityFixtures(tb testing.TB, db *bun.DB, ids ...int64) {
 			Where("active_group_id IN (SELECT id FROM active.groups WHERE group_id = ?)", id),
 			"active.visits (cascade)")
 
-		// Delete from active.groups by direct ID or by reference
+		// Delete from active.groups by direct ID or by reference.
+		// room_id reference is required so facilities.rooms can be deleted
+		// without tripping fk_active_groups_room_tenant.
 		cleanupDelete(tb, db.NewDelete().
 			TableExpr("active.groups").
-			Where("id = ? OR group_id = ? OR device_id = ?", id, id, id),
+			Where("id = ? OR group_id = ? OR device_id = ? OR room_id = ?", id, id, id, id),
 			"active.groups")
 
 		// ========================================
@@ -425,10 +427,12 @@ func CleanupActivityFixtures(tb testing.TB, db *bun.DB, ids ...int64) {
 			Where("activity_group_id = ?", id),
 			"activities.student_enrollments")
 
-		// Delete from activities.groups by ID or by category_id (to handle FK constraint)
+		// Delete from activities.groups by ID, by category_id, or by created_by.
+		// created_by reference is required so users.staff can be deleted without
+		// tripping fk_activity_groups_created_by_tenant.
 		cleanupDelete(tb, db.NewDelete().
 			TableExpr("activities.groups").
-			Where("id = ? OR category_id = ?", id, id),
+			Where("id = ? OR category_id = ? OR created_by = ?", id, id, id),
 			"activities.groups")
 
 		// Delete from activities.categories (now safe after groups referencing them are deleted)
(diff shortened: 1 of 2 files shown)


yungweng:
@creintjes pls fix die tests, danach kann ich mergen

creintjes:
CI flagged 3 \`TestSchoolCheckin_*\` failures on the first run (rerun was green, hence \`test / test-backend\` job is success but \`Backend Test Results\` reporter shows fail). I reproduced locally with \`-count=2\` and root-caused it.

## Root cause

Migration 1.7.5 inserts \`WEB-MANUAL-001\` as the very first row in \`iot.devices\`. In a freshly-migrated DB it gets \`id=1\` (BIGSERIAL starts at 1). The web-checkin handler falls back to this device when no kiosk \`deviceID\` is provided.

\`CleanupActivityFixtures\` deletes \`iot.devices WHERE id = ?\` for every passed fixture id. As soon as anything passes \`id=1\` — even transiently, even from an unrelated chain — the system device is gone, and subsequent web check-ins trip \`fk_attendance_device_tenant\`. \`resolveDeviceIDForAttendance\` logs the warning and falls through to \`deviceID=0\`, which the FK rejects.

This race was always latent; my cleanup-completeness fix makes it more reliable to trigger because cleanups now actually run end-to-end.

## Fix pushed (\`65e34259\`)

Same convention we already use for \`rooms.id=1\`, \`schools.id=1\`, and the new \`staff.id=1\`: exclude the system device by \`device_id\` at the cleanup site.

\`\`\`go
cleanupDelete(tb, db.NewDelete().
    TableExpr("iot.devices").
    Where(whereIDEquals, id).
    Where("device_id != ?", "WEB-MANUAL-001"),  // never delete system device
    "iot.devices")
\`\`\`

## Verification

\`\`\`bash
docker compose --profile test down postgres-test -v
docker compose --profile test up -d postgres-test
APP_ENV=test docker compose run server ./main migrate
go test -p 1 -count=2 ./...   # full suite, twice in same DB, all green
\`\`\`

After both runs: \`SELECT id FROM iot.devices WHERE device_id='WEB-MANUAL-001'\` still returns id=1. No leaks, no FK violations.

Re-running CI; should be clean now.

creintjes:
@yungweng Ist das normal, dass der eine Test seit >50min in der Warteschlange ist?

yungweng:
@creintjes perhaps ein Blacksmith issue. Hab das jetzt neu gestartet

creintjes:
Alles klar, danke! Ich kümmer mich um den gefailten Test

PR #1337, opened 2026-04-28
feat(database/rooms): customizable badge colors per room

### LINKED ISSUE #1324: Farbcodierung für Raum-Badges customisierbar
## Idee
Aktuell sind alle Raum-Badges blau (Ausnahme: "unterwegs" = lila). Pflegekräfte sollen pro Badge-Status individuelle Farben definieren können (Color Picker).

## Begründung
Die einheitliche Färbung erschwert die schnelle Orientierung auf einen Blick. Individuelle Farben pro Status (z.B. "im Haus" grün, "im Hof" gelb, "unterwegs" lila) verbessern die Übersicht in der Schul-OGS deutlich.

## Akzeptanzkriterien
- [ ] Color Picker pro Badge-Status in den Schuleinstellungen
- [ ] Änderung wirkt sich live auf alle Raumansichten und Listen aus
- [ ] Default-Farbwerte bleiben erhalten falls nicht überschrieben

## Quelle
Florian Lüttgenau, Slack-Post am 2026-04-28 10:08
Slack-Vorschlag: Tristan zuständig

theitger:
Closes #1324. Replaces the all-blue room badge with admin-pickable hex codes. Status colors (Schulhof, Unterwegs, Zuhause, Krank, Entschuldigt) and the "own group room" green keep their meaning — only the generic `OTHER_ROOM` blue fallback is replaced when a room has a color set. System rooms (Schulhof, WC) cannot be re-coloured.

## Backend

- Migration `1.15.44` backs up every row carrying the legacy `#4F46E5` value into `audit.room_color_migration_backup` and NULLs it on `facilities.rooms`. That hex was never user-chosen — it came from a `transformBeforeSubmit` bug that stamped it onto every save, invisible until now because the badge ignored `room.color`. Migration then adds a partial `UNIQUE(tenant_id, lower(color)) WHERE color IS NOT NULL` index. Rollback drops the index but does not restore cleared colors.
- `reservedRoomColors` blocks the status palette plus `#4F46E5`. A drift test parses `LOCATION_COLORS` from the frontend and fails CI if the two diverge.
- System rooms reject color writes; edits that omit the `color` field preserve the existing value (so a Schulhof capacity bump still works on rows carrying leftover legacy color).
- Validation centralised in the service layer; model sentinels translated to 400 / 409 with German messages via `ErrorRenderer`. Color-conflict path matches the unique-constraint name via `pgdriver.Error.Field('n')` so `ErrColorAlreadyInUse` is not shadowed by the generic `ErrDuplicateRoom`.
- `StudentResponse.current_room_color` and `RoomSimple.color` carry the value into student / visit payloads.

## Frontend

- New `RoomColorField` (circle preview + reset button) on Database → Rooms; hidden for system rooms.
- `getLocationColor` priority: own group room green → per-room hex → `OTHER_ROOM` blue. Status colors are unchanged.
- `useTenantMutateMatching` invalidates room-derived caches after a save; the substring list lives in `lib/swr/room-derived-caches.ts` as a single source of truth.
- `current_room_color` threaded through active-supervisions, OGS-Groups, and students-search via `mapStudentResponse` and the existing BFF passthrough routes.

## Test plan

- [ ] `cd backend && go test ./...` and `cd frontend && pnpm run check`
- [ ] Apply migration `1.15.44` on a DB with `#4F46E5` rows; verify backup table is populated and index `uniq_facilities_rooms_tenant_color` exists
- [ ] Pick a custom hex on a non-system room → badge updates on Database, Active-Supervisions, and OGS-Groups
- [ ] Reserved hex (e.g. `#83CD2D`) → 400 with German message
- [ ] Same hex on two rooms in one school → 409 with German message
- [ ] Color write on Schulhof / WC rejected; capacity-only edit on the same row still succeeds

### DIFF
--- a/backend/api/active/converters.go
+++ b/backend/api/active/converters.go
@@ -62,8 +62,9 @@ func newActiveGroupResponse(group *active.Group) ActiveGroupResponse {
 	// Add room info if available
 	if group.Room != nil {
 		response.Room = &RoomSimple{
-			ID:   group.Room.ID,
-			Name: group.Room.Name,
+			ID:    group.Room.ID,
+			Name:  group.Room.Name,
+			Color: group.Room.Color,
 		}
 	}
 
--- a/backend/api/active/converters_color_test.go
+++ b/backend/api/active/converters_color_test.go
@@ -1,0 +1,83 @@
+package active
+
+import (
+	"encoding/json"
+	"testing"
+	"time"
+
+	activeModels "github.com/moto-nrw/project-phoenix/models/active"
+	modelBase "github.com/moto-nrw/project-phoenix/models/base"
+	"github.com/moto-nrw/project-phoenix/models/facilities"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TestNewActiveGroupResponse_RoomColorPropagation guards the most fragile
+// link in the badge-color pipeline: the RoomSimple struct used by every
+// active-group response. The frontend BFFs decode this JSON to populate
+// `current_room_color` for badges. If anyone:
+//   - removes Color from RoomSimple,
+//   - forgets to set it in newActiveGroupResponse, or
+//   - changes the JSON tag from "color" to something else,
+//
+// every per-room badge color silently disappears with no compile error.
+//
+// This test marshals the response struct and inspects the raw JSON to catch
+// all three regressions in one shot.
+func TestNewActiveGroupResponse_RoomColorPropagation(t *testing.T) {
+	now := time.Now()
+	color := "#A3D977"
+
+	t.Run("color is included in JSON when room has color set", func(t *testing.T) {
+		group := &activeModels.Group{
+			Model:     modelBase.Model{ID: 1, CreatedAt: now, UpdatedAt: now},
+			RoomID:    10,
+			StartTime: now,
+			Room: &facilities.Room{
+				Model: modelBase.Model{ID: 10},
+				Name:  "Werkraum",
+				Color: &color,
+			},
+		}
+
+		resp := newActiveGroupResponse(group)
+		require.NotNil(t, resp.Room, "Room must be populated when group.Room is set")
+
+		// Direct field check — catches "Color field removed" regression.
+		require.NotNil(t, resp.Room.Color)
+		assert.Equal(t, "#A3D977", *resp.Room.Color)
+
+		// JSON-shape check — catches "json tag renamed" regression. The
+		// frontend BFF reads this exact key, so a tag rename would silently
+		// drop the value.
+		raw, err := json.Marshal(resp.Room)
+		require.NoError(t, err)
+		assert.JSONEq(t, `{"id":10,"name":"Werkraum","color":"#A3D977"}`, string(raw))
+	})
+
+	t.Run("color is omitted from JSON when room has no color", func(t *testing.T) {
+		// When color is nil the response must still be valid JSON without
+		// a "color" key — the omitempty tag earns its keep here. Without
+		// it, the field would serialize as "color":null, which the
+		// frontend would treat as an explicit cleared value (still falls
+		// back to blue, but it's a noisier wire payload).
+		group := &activeModels.Group{
+			Model:     modelBase.Model{ID: 1},
+			RoomID:    10,
+			StartTime: now,
+			Room: &facilities.Room{
+				Model: modelBase.Model{ID: 10},
+				Name:  "Bibliothek",
+			},
+		}
+
+		resp := newActiveGroupResponse(group)
+		require.NotNil(t, resp.Room)
+		assert.Nil(t, resp.Room.Color)
+
+		raw, err := json.Marshal(resp.Room)
+		require.NoError(t, err)
+		assert.JSONEq(t, `{"id":10,"name":"Bibliothek"}`, string(raw),
+			"omitempty must drop the color key when not set")
+	})
+}
--- a/backend/api/active/types.go
+++ b/backend/api/active/types.go
@@ -63,10 +63,16 @@ type GroupSupervisorSimple struct {
 	Role    string `json:"role,omitempty"`
 }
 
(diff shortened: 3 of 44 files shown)


yungweng:
Alright, sitting down with a coffee for this one. Honestly some of the cleanest work to land on this repo in a while — backup table on the migration, asymmetric rollback semantics, drift test that parses the *actual* TS source, omitempty regression guard via raw JSON marshalling. You keep raising the bar and the rest of us keep tripping over it. The one good dev around here, evidently.

That said, two things will bite us at deploy time, so I have to put on the grumpy reviewer hat for a minute.

## Must fix

**1. The legacy `#FFFFFF` default will break your index creation.** Migration `001001001_facilities_rooms.go:60` shipped `color TEXT NOT NULL DEFAULT '#FFFFFF'`, and `001001005` dropped the default but never nulled the existing rows. Any tenant with two rooms created before 1.1.5 has duplicate `#FFFFFF` values, and your `CREATE UNIQUE INDEX ... WHERE color IS NOT NULL` will refuse to create on those tenants. I love that you handled `#4F46E5` so carefully — please give `#FFFFFF` the same treatment (back up + null) before the index goes on. Or generalize: dedupe any per-tenant color collisions before the index lands.

**2. Shorthand hex bypasses the uniqueness rule.** `Validate()` happily accepts `#RGB` and stores it verbatim. Room A picks `#ABC`, Room B picks `#AABBCC`, both render as the exact same CSS color, and `LOWER(color)` thinks they're different strings. Native picker always emits 6-digit so the UI is fine, but a direct API call slips right through. Easy patch: expand shorthand to `#RRGGBB` right after the regex match, before the upper-case normalisation.

## Should fix (non-blocking but worth a follow-up)

- The `RoomColorField` picker preview is bound to `LOCATION_COLORS.OTHER_ROOM`, which is reserved. Native `<input type=\"color\">` only fires `onChange` on actual movement so this rarely bites in practice — but a deliberate \"yes I want exactly that blue\" pick produces a confusing German 400. Consider seeding the preview with a non-reserved neutral.
- `ErrorRenderer` has no direct unit tests for the new branches. The 400/409/403/500 mapping is exactly the kind of contract a 20-line table test should pin.
- `exposedReservedHexes` in the drift test maintains its own hardcoded mirror of the palette half. Adding a future status color now needs three edits. Tiny `PaletteReservedHexes()` accessor on `room_colors.go` would let the test interrogate the source of truth instead of duplicating it.
- `vi.fn(() => vi.fn())` in `database/rooms/page.test.tsx` returns a fresh function per call — fine for current assertions, but if anything starts using callback identity as a deps trigger, the mock will hide it. Stable reference is a one-liner.

## Nits

- Comment in `room_colors_test.go` (\"this test won't catch that drift\") is stale — your drift test catches it bidirectionally via `setDiff`.
- `ErrColorReserved` German message says \"reserviert für Statusbadges\" but `#4F46E5` is reserved for migration-restore correctness, not because it's a status badge. Either a separate sentinel with a tailored message, or accept the trade-off and move on.
- `testpkg.CleanupActivityFixtures` for room cleanup reads weird at the call site — possibly a generic helper, but a `CleanupRoom` alias would make the intent obvious.
- The `extractLocationColorsHexes` regex is brittle by design. An anchor comment in `location-helper.ts` (`// drift-test-anchor`) would make the cross-codebase coupling visible from the frontend side too.
- `current_room_color` JSDoc has the maintenance note pointing at `room-derived-caches.ts` — same hint would earn its keep on `active-supervisions/page.tsx` and `ogs-groups/page.tsx`.

## What I genuinely loved

- `pgdriver.Error.Field('n')` matching the constraint name so `ErrDuplicateRoom` doesn't shadow the more specific `ErrColorAlreadyInUse`. Chef's kiss.
- The defense-in-depth in `ErrorRenderer` (sentinel path *and* the bare `IsValidationError` fallback) so a missed `FacilitiesError` wrap doesn't quietly turn validation errors into 500s.
- Single `INSERT ... SELECT` for the backup snapshot to avoid the enumerate-then-clear race.
- Centralising `ROOM_DERIVED_CACHE_KEY_FRAGMENTS` into one file with an actual maintenance contract attached, instead of inlining the substring list in the page.

Fix the two `Must fix`es and this is a clean ship. Everything else can be follow-ups.

theitger:
@yungweng Thanks for the careful read — both must-fixes are in as `18fdedc1`, plus a follow-up `ebd37896` to clear a migration version collision that landed on `development` in the meantime.

**1. `#FFFFFF` legacy default (migration 1.15.45)**

Same treatment as `#4F46E5`, since both share the "never user-chosen" provenance:

- Backup `INSERT … SELECT` widened to `WHERE LOWER(color) IN (LOWER('#4F46E5'), LOWER('#FFFFFF'))`
- `UPDATE … SET color = NULL` widened to the same predicate
- Backup table schema unchanged — the `color` column preserves the original hex per row, so a manual restore can tell `#4F46E5` rows apart from `#FFFFFF` rows
- godoc + log lines updated to call out both legacy defaults and cite migrations 1.1.1 (NOT NULL DEFAULT) and 1.1.5 (dropped default but never nulled rows) as the sources

**2. `#RGB` shorthand bypass (`Room.Validate()`)**

Expansion happens immediately after the regex match, before both the upper-case normalisation and the `IsReservedRoomColor` check:

```go
if len(*r.Color) == 4 {
    s := *r.Color
    expanded := string([]byte{'#', s[1], s[1], s[2], s[2], s[3], s[3]})
    r.Color = &expanded
}
```

So `#abc` and `#AABBCC` now both persist as `#AABBCC`, and the reserved-color check sees the canonical 6-digit form too — any future reserved hex of repeating-pair shape would still trip on a shorthand input. New unit test pins `#abc → #AABBCC` and a second test covers the missing-`#` path (`abc` → prefix → expand → uppercase) so the canonicalisation order is locked in.

Thanks again for the chef's-kiss-level review.

yungweng (approved):
lgtm;just believe

PR #1345, opened 2026-04-29
fix(iot): return 409 Conflict on duplicate active visit

### LINKED ISSUE #844: Duplicate active visit returns HTTP 500 instead of 409 Conflict
## Context

Production log analysis 2026-02-05, 14:38 CET.

## Problem

`POST /api/iot/checkin` returns HTTP 500 when a student already has an active visit (`student_id: 231`). This should be a **409 Conflict** — the request itself is valid, but it conflicts with existing state.

A 500 signals a server error to the IoT device, which may trigger retries or misleading error displays. A 409 clearly communicates "already checked in" and allows PyrePortal to handle it gracefully.

## Expected behavior

- Return **409 Conflict** with a clear error message (e.g. `"student already has an active visit"`)
- Include the existing visit ID in the response so the client can display context

## Steps to investigate

1. Trace the checkin handler → service → repository flow for duplicate visit detection
2. Identify where the error originates (likely a unique constraint violation or service-level check)
3. Map the error to a proper 409 response instead of letting it bubble up as 500
4. Verify PyrePortal handles 409 correctly (or create follow-up issue)

## Production log excerpt

```
HTTP 500 — POST /api/iot/checkin
Cause: student already has an active visit (student_id: 231)
```

theitger:
Closes #844. Companion kiosk PR: moto-nrw/PyrePortal#332.

## Problem
`POST /api/iot/checkin` returned HTTP 500 when a student already had an open
`active.visits` row (production log, student_id 231). 500 signals a server
error, so PyrePortal showed a generic failure and risked retry storms; the
condition is in fact a state conflict and should be a 409.

### IoT checkin path (`api/iot/checkin`)
- Map `services/active.ErrStudentAlreadyActive` to **HTTP 409** with a
  structured body so the kiosk can render "Bereits angemeldet in <Raum>":
  ```json
  {
    "error": "student already has an active visit",
    "code": "STUDENT_ALREADY_ACTIVE",
    "details": {
      "student_id": 231,
      "existing_visit_id": 9123,
      "entry_time": "2026-02-05T13:14:22Z",
      "room_id": 12,
      "room_name": "Lernraum 2"
    }
  }
  ```
- `entry_time` is `*time.Time` so the degraded path (existing-visit lookup
  fails) drops the field instead of marshalling `0001-01-01T00:00:00Z`.

### Admin visits route (`api/active`)
- `ErrStudentAlreadyActive` was previously classified as 400 by
  `api/active.ErrorRenderer`. With the DB-level translation introduced
  below, every `CreateVisit` caller can now hit it, so the admin route
  would have surfaced the same state conflict as 400 while the IoT path
  returned 409. Both endpoints now agree on **409**.

### DB-level race close (migration 1.15.46)
- Adds partial unique index `uniq_active_visits_open_per_student` on
  `active.visits (tenant_id, student_id) WHERE exit_time IS NULL` so two
  concurrent checkins can't both pass the application-level idempotency
  check and double-insert.
- `services/active.CreateVisit` translates the resulting `pgerrcode 23505`
  on that constraint to `ErrStudentAlreadyActive`, so the read-then-write
  race and the steady-state duplicate hit the same 409 response.
- Migration also dedupes any existing duplicates (keeps the most recent
  open row per student) before the index is created, so it's safe to apply
  on environments that already drifted.

## Tests
- `TestDeviceCheckin_DuplicateActiveVisit_AppLevelPath_Returns409WithRoomDetails`
  — handler-level 409 + structured body for the application path.
- `TestErrorStudentAlreadyActive_*` — JSON shape on the full path and the
  degraded path (only `student_id` survives when the existing-visit lookup
  failed).
- `TestIsDuplicateActiveVisitViolation` — inserts an open visit, attempts a
  second open insert, and asserts the real
  `*base.DatabaseError → bun → pgdriver.Error` chain is recognised by the
  detection helper. Exercises the migration's index end-to-end.
- `TestErrorRenderer_StudentAlreadyActiveConflict` — guards the admin-route
  409 mapping so a future reshuffle can't silently regress it back to 400.

## Notes
- Migration version was bumped 1.15.45 → 1.15.46 because development picked
  up an unrelated `rooms_color_unique` migration on the same slot mid-flight
  (`MigrationRegistry` keys on the version string and silently overwrites
  on collision).
- Cross-repo contract: PyrePortal must match on `code: STUDENT_ALREADY_ACTIVE`
  and read `details.room_name` / `details.student_id` — see companion PR.

### DIFF
--- a/backend/api/active/active_internal_test.go
+++ b/backend/api/active/active_internal_test.go
@@ -128,11 +128,10 @@ func TestErrorRenderer_AllBadRequestErrors(t *testing.T) {
 			err:          &activeService.ActiveError{Op: "test", Err: activeService.ErrStudentAlreadyInGroup},
 			expectedText: "Student Already In Group",
 		},
-		{
-			name:         "ErrStudentAlreadyActive",
-			err:          &activeService.ActiveError{Op: "test", Err: activeService.ErrStudentAlreadyActive},
-			expectedText: "Student Already Has Active Visit",
-		},
+		// ErrStudentAlreadyActive intentionally absent — it maps to
+		// 409 Conflict (see TestErrorRenderer_StudentAlreadyActive in
+		// handlers_unit_test.go and TestErrorRenderer_StudentAlreadyActiveConflict
+		// in errors_test.go) per the Issue #844 review fix.
 		{
 			name:         "ErrStaffAlreadySupervising",
 			err:          &activeService.ActiveError{Op: "test", Err: activeService.ErrStaffAlreadySupervising},
--- a/backend/api/active/errors.go
+++ b/backend/api/active/errors.go
@@ -85,7 +85,17 @@ func ErrorRenderer(err error) render.Renderer {
 		renderer.StatusText = "Student Already In Group"
 
 	case errors.Is(err, activeSvc.ErrStudentAlreadyActive):
-		renderer.HTTPStatusCode = http.StatusBadRequest
+		// 409 Conflict — the request is well-formed but conflicts with
+		// existing state (the student already has an open visit). Issue
+		// #844 added a DB-level partial unique index on active.visits;
+		// the active service now translates the resulting 23505 error
+		// to ErrStudentAlreadyActive for ALL CreateVisit callers, not
+		// just the IoT checkin flow. Without this 409 mapping, the
+		// admin POST /active/visits route would surface concurrent
+		// duplicate-creation as 400 Bad Request, contradicting both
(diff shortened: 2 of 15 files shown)


yungweng (approved):
lgtm;good one

yungweng (changes requested):
bitte erst in pyreportal den fix machen. Bitte keine Hardcoded Timeouts einbauen, das ist aus UX sicht grob fatal.

yungweng (approved):
lgtm

PR #1355, opened 2026-05-01
feat(facilities): accept WC and Toilette as toilet room aliases (#1184)

### LINKED ISSUE #1184: Accept both "WC" and "Toilette" as toilet room aliases across Phoenix and PyrePortal
## Summary
The toilet special-room flow currently depends on the room being named exactly `WC`. This is too strict for setup and documentation, where users may create the room as `Toilette`.

We should support both `WC` and `Toilette` as valid aliases while preserving backwards compatibility for existing tenants.

## Problem
Current behavior is split across both repos:
- Phoenix backend treats the toilet special room as `WC` only.
- PyrePortal only shows the toilet destination when it finds a room named exactly `WC`.
- A manually created room named `Toilette` is therefore not recognized end to end.

## Desired behavior
- Existing tenants using `WC` continue to work unchanged.
- Tenants with a manually created room named `Toilette` also work.
- If neither room exists, auto-creation creates exactly one canonical room, not duplicates.
- The UI continues to display `Toilette`.

### Phoenix
- Accept both `WC` and `Toilette` when identifying the toilet special room.
- In the WC auto-create flow, first search for existing aliases before creating a new room.
- Avoid creating duplicate special rooms when one alias already exists.
- Keep activity/category compatibility intact.

### PyrePortal
- Detect both `WC` and `Toilette` when deciding whether to show the toilet destination button.
- Continue displaying the user-facing label as `Toilette`.

## Acceptance criteria
- A room named `WC` works exactly as today.
- A room named `Toilette` also enables the toilet flow.
- PyrePortal shows the `Toilette` destination for either room name.
- Phoenix reuses either existing alias before auto-creating anything.
- When neither alias exists, only one canonical room is auto-created.
- No duplicate toilet special rooms are created by alias handling.
- Relevant tests are updated in both repos.

## Notes
This is a cross-repo change:
- `moto-nrw/project-phoenix`
- `moto-nrw/PyrePortal`

It is fine to track it in Phoenix, but implementation needs coordinated changes in both repositories.

theitger:
Companion PR: https://github.com/moto-nrw/PyrePortal/pull/336
Closes #1184

## Summary

- `IsWCRoomName(name)` accepts both `WC` (canonical, auto-created) and `Toilette` (manually-created alias). `IsSystemRoomName` delegates to it, so rename/delete/color protection automatically covers `Toilette` rooms — no separate guard needed.
- New `Service.FindToiletRoom(ctx, excludeRoomID)` resolves the toilet special-room by iterating canonical-first (`WC` then `Toilette`). Used by the auto-create flow in `api/iot/checkin/wc.go` and `services/facilities/wc_service.go` so an existing alias is reused instead of producing duplicates.
- **Exact-case contract across all layers** (per #1184 review): only the names `"WC"` and `"Toilette"` count as toilet system rooms. `FindToiletRoom` re-filters the result of the case-insensitive `RoomRepository.FindByName` with `IsWCRoomName` so non-canonical case variants (e.g. lowercase `wc`) are skipped. This aligns `FindToiletRoom` with `IsWCRoomName`, `IsSystemRoomName`, the rename/delete guards, the IoT scan-fallback in `api/iot/checkin/workflow.go`, and the partial unique index from migration 1.15.48 — all exact-case.
- `CreateRoom` and `UpdateRoom` reject creating/renaming to a second alias when one already exists for the tenant, with a DB-level partial unique index `WHERE name IN ('WC','Toilette')` (migration 1.15.48) closing the TOCTOU race the app guard can't reach.
- Migration 1.15.48 also archives pre-existing duplicates per tenant — picks a canonical winner (most active groups, then `WC` over `Toilette`, then oldest id), renames losers to `<name> (archiviert v1.15.48 id=<id>)`, and writes originals to `audit.wc_alias_migration_backup` for manual restore.
- Frontend `SYSTEM_ROOM_NAMES` in `room-helpers.ts` adds `Toilette` so the rooms admin UI treats it as a protected system room (sync note links the three-way invariant: backend constants, frontend helper, PyrePortal `WC_ROOM_ALIASES`).
- Activity / category logic untouched — only room-name detection broadens, the WC activity name remains `WCActivityName`.

## Behavior matrix

| Tenant state | Before | After |
|---|---|---|
| Has room `WC` | Toilet flow works | Toilet flow works (unchanged) |
| Has room `Toilette` (manually created) | Auto-create makes a duplicate `WC` room | Existing `Toilette` is reused |
| Has neither | Auto-creates one canonical `WC` | Same |
| Has both (legacy / accidental) | N/A | Migration 1.15.48 picks one and archives the loser |
| Admin tries to create `Toilette` while `WC` exists | Allowed (duplicate) | Rejected with `ErrDuplicateRoom` (app guard + DB index) |
| Has only lowercase `wc` (no canonical) | Toilet flow worked (silent adoption) | **Toilet flow returns an error until admin renames the room to `WC`** |

## Behavioral regression to flag

Tenants that have *only* a lowercase `wc` room (no canonical `WC` or `Toilette`) lose the kiosk toilet flow until an admin renames the room. This is the cost of picking the exact-case contract from yungweng's review — the previous "silent adoption" was the bug, because the adopted room was unprotected against rename/delete and invisible to other layers. Stuck state surfaces as a clear error, not silent corruption. Worth a heads-up for staging operators on rollout.

## Out of scope

- The case-insensitive lookup in `RoomRepository.FindByName` (`LOWER(name) = LOWER(?)`) is pre-existing and intentionally preserved — `CreateRoom`'s duplicate-name guard relies on it. `FindToiletRoom` enforces exact-case at the call site via `IsWCRoomName` re-check rather than tightening `FindByName` itself.
- Pre-existing failures in `TestRepairPickupScheduleTenantIDs*` (migration 1.15.39 tests) are unrelated to this branch.

## Test plan

- [ ] `cd backend && go test ./constants/... ./services/facilities/... ./api/iot/checkin/... ./database/migrations/... -run "WC|Toilet|System"` — all alias tests green (includes new `TestFindToiletRoom_SkipsLowercaseWCRoom`, `TestEnsureWCRoom_IgnoresLowercaseWCRoom`, `TestWCService_ensureWCRoom_IgnoresLowercaseWCRoom`)
- [ ] `cd backend && go test ./test/ -run TestHermeticTestPatterns -v` — hermetic check green
- [ ] `cd backend && ./main migrate validate` — dependency graph valid
- [ ] `cd frontend && pnpm run check` — no warnings
- [ ] On a staging tenant that *only* has a `WC` room: kiosk Toilette flow still works, auto-create still produces `WC`
- [ ] On a staging tenant that *only* has a `Toilette` room: kiosk Toilette flow works, no duplicate `WC` is created on first scan
- [ ] On a staging tenant with neither: kiosk Toilette flow auto-creates `WC` (canonical)
- [ ] On a staging tenant with *only* a lowercase `wc` room: kiosk Toilette flow surfaces an error (admin must rename to `WC` to unblock)
- [ ] Admin UI: deleting/renaming/recoloring a `Toilette` room is blocked with the existing system-room message
- [ ] Admin UI: creating a second toilet alias while one exists returns `ErrDuplicateRoom`
- [ ] Run the migration on a staging snapshot containing pre-existing duplicates — losers archived, `audit.wc_alias_migration_backup` populated, partial unique index built

## Cross-repo coordination

Phoenix is backwards-compatible with the old kiosk (no behavior change for tenants currently on `WC`), so this can land first. The companion PyrePortal PR adds the kiosk-side alias detection so `Toilette`-named rooms light up the toilet button.

### DIFF
--- a/backend/api/iot/checkin/wc.go
+++ b/backend/api/iot/checkin/wc.go
@@ -2,6 +2,7 @@ package checkin
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"log/slog"
 
@@ -9,17 +10,21 @@ import (
 	"github.com/moto-nrw/project-phoenix/models/activities"
 	"github.com/moto-nrw/project-phoenix/models/base"
 	"github.com/moto-nrw/project-phoenix/models/facilities"
+	facilitiesSvc "github.com/moto-nrw/project-phoenix/services/facilities"
 )
 
 // ensureWCRoom finds or creates the WC room
 func (rs *Resource) ensureWCRoom(ctx context.Context) (*facilities.Room, error) {
-	// Try to find existing WC room
-	room, err := rs.FacilityService.FindRoomByName(ctx, constants.WCRoomName)
+	room, err := rs.FacilityService.FindToiletRoom(ctx, 0)
 	if err == nil && room != nil {
 		rs.getLogger().DebugContext(ctx, "found existing WC room",
 			slog.Int64("room_id", room.ID),
+			slog.String("room_name", room.Name),
 		)
 		return room, nil
+	}
+	if err != nil && !errors.Is(err, facilitiesSvc.ErrRoomNotFound) {
+		return nil, fmt.Errorf("failed to look up WC room: %w", err)
 	}
 
 	// Room not found - create it
@@ -37,9 +42,8 @@ func (rs *Resource) ensureWCRoom(ctx context.Context) (*facilities.Room, error) 
 	}
 
 	if err := rs.FacilityService.CreateRoom(ctx, newRoom); err != nil {
-		// Retry: concurrent request may have created it
-		room, retryErr := rs.FacilityService.FindRoomByName(ctx, constants.WCRoomName)
-		if retryErr == nil && room != nil {
+		// Retry: concurrent request may have created one of the accepted aliases
+		if room, retryErr := rs.FacilityService.FindToiletRoom(ctx, 0); retryErr == nil && room != nil {
 			return room, nil
 		}
 		return nil, fmt.Errorf("failed to auto-create WC room: %w", err)
--- a/backend/api/iot/checkin/wc_room_alias_integration_test.go
+++ b/backend/api/iot/checkin/wc_room_alias_integration_test.go
@@ -1,0 +1,170 @@
+package checkin_test
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"testing"
+	"time"
+
+	"github.com/go-chi/chi/v5"
+	"github.com/moto-nrw/project-phoenix/api/testutil"
+	"github.com/moto-nrw/project-phoenix/constants"
+	"github.com/moto-nrw/project-phoenix/models/active"
+	"github.com/moto-nrw/project-phoenix/models/facilities"
+	testpkg "github.com/moto-nrw/project-phoenix/test"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/uptrace/bun"
+)
+
+func cleanupWCRoomAliasIntegrationArtifacts(t *testing.T, db *bun.DB) {
+	t.Helper()
+
+	dbCtx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	stmts := []string{
+		fmt.Sprintf(`DELETE FROM active.attendance WHERE visit_id IN (SELECT v.id FROM active.visits v JOIN active.groups ag ON ag.id = v.active_group_id JOIN facilities.rooms r ON r.id = ag.room_id WHERE r.name IN ('%s', '%s'))`, constants.WCRoomName, constants.WCRoomAliasName),
+		fmt.Sprintf(`DELETE FROM active.visits WHERE active_group_id IN (SELECT ag.id FROM active.groups ag JOIN facilities.rooms r ON r.id = ag.room_id WHERE r.name IN ('%s', '%s'))`, constants.WCRoomName, constants.WCRoomAliasName),
+		fmt.Sprintf(`DELETE FROM active.group_supervisors WHERE group_id IN (SELECT ag.id FROM active.groups ag JOIN facilities.rooms r ON r.id = ag.room_id WHERE r.name IN ('%s', '%s'))`, constants.WCRoomName, constants.WCRoomAliasName),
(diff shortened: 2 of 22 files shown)


yungweng:
wenigstens einer der den Tag der Arbeit gebührend feiert!

yungweng:
@theitger thanks for digging into this, especially on a holiday. I really appreciate the care here.

I do think this needs to be fixed before merge, but I’d frame the issue slightly broader than “make the lookup exact-case”: right now the WC-room identity contract is inconsistent across layers.

`FindToiletRoom` can return lowercase/non-canonical rooms because `FindByName` is case-insensitive, but `IsWCRoomName`, rename/delete protection, the IoT special-room path, and the partial unique index only treat exact `"WC"` / `"Toilette"` as system aliases. That means a lowercase `wc` or `toilette` can be reused as WC infrastructure without being protected or recognized consistently later.

So we should pick one contract and make all layers follow it:

- exact-case only: change the alias lookup to exact matching; or
- backward-compatible lowercase reuse: normalize/protect the matched room consistently everywhere.

Either direction is fine, but shipping the current split behavior seems risky.

theitger:
@yungweng went with **Option A (exact-case everywhere)**. Pushed `c1a89263`:

- `FindToiletRoom` now re-filters `FindByName` results through `IsWCRoomName`, so case variants (`wc`, `toilette`, …) are skipped instead of silently adopted.
- All five layers now agree: `IsWCRoomName`, `IsSystemRoomName` rename/delete guards, IoT scan-fallback in `workflow.go`, partial unique index from migration 1.15.48, and `FindToiletRoom`.
- `FindByName` itself stays case-insensitive (the `CreateRoom` duplicate-name guard relies on it) — the contract is enforced at the call site.
- Tests flipped: lowercase `wc` is no longer reused; the duplicate-name collision surfaces as an error. Added `TestFindToiletRoom_SkipsLowercaseWCRoom` for the lookup primitive.

**Heads-up before rollout**: any tenant with a lowercase `wc`/`toilette` room loses the kiosk toilet flow until renamed. A quick

```sql
SELECT tenant_id, id, name FROM facilities.rooms
WHERE LOWER(name) IN ('wc','toilette') AND name NOT IN ('WC','Toilette');
```

on staging (and again on prod before that deploy) will surface affected tenants. The previous silent adoption was the bug — adopted rooms were unprotected and invisible to the IoT path.

Ready for another look 🙏

yungweng (approved):
gute arbeit, wird schon passen

PR #1388, opened 2026-05-08
fix(students): support full-name search on database page

theitger:
Closes #1387

## Summary

Fixes the student search on `Datenbank > Schüler` so searches spanning first and last name work again.

Previously the filter only checked `first_name` and `second_name` separately, so inputs like `Max Mu` or `Max Mustermann` returned no results. The filter now also matches combined name variants and adds a regression test for that case.

## Validation

- `pnpm --dir ./frontend test:run 'src/app/[tenant]/(protected)/database/students/page.test.tsx'`
- `pnpm --dir ./frontend check`


yungweng (approved):
looks good to me, i believe

PR #1385, opened 2026-05-08
feat(students): add tenant-scoped student photos with consent and avatars

theitger:
## Summary

This PR adds tenant-scoped student photos with parental consent, authenticated photo serving, and avatar rendering across student-related staff views.

When `operations.student_photos_enabled` is enabled for a school, staff can upload, replace, and remove student photos in the student profile. Photos are only available when consent is present, follow existing student access rules, and fall back to initials when no image is available.

## Included

- per-tenant `operations.student_photos_enabled` setting
- student photo upload / delete / authenticated serve routes
- parental consent handling with audit metadata
- automatic photo removal on consent withdrawal
- automatic purge of stored photos when the feature is disabled
- avatar rendering in student cards, search, room/detail, and related views
- tenant/tab invalidation to keep open views in sync after settings or photo changes
- race-safe cleanup and post-commit file deletion/broadcast handling

## Why this PR is broad

Student photos affect more than the student form itself. To behave correctly, the feature also needs settings integration, access control, authenticated file serving, cache invalidation, realtime updates, and cleanup when consent or feature state changes.

## Testing

- backend tests added for photo routes, consent transitions, purge flow, response shaping, SSE, and after-commit behavior
- frontend tests added for avatar rendering, photo UI, proxy routes, tenant invalidation, SSE handling, and student API helpers
- manually verified enable/disable flow, upload/remove flow, consent behavior, and live updates across open tabs

Closes #1370

PR #1391, opened 2026-05-10
refactor: settings side-effects registry + single SWR consumer for the schema

theitger:
Standalone refactor extracted from the (now closed) #1385 student-photo branch in response to @yungweng's review. First of two PRs — the photo feature itself follows in a stacked PR once this lands.

Addresses two of @yungweng's review points on #1385:

> _"The settings side effects are wired in API bootstrap — backend/api/base.go:307. [...] Putting that in `initializeAPIResources` makes the API constructor a hidden business-process registry."_

> _"Tab B can show stale settings and resubmit old values. That is not theoretical."_

## Summary

**Backend** (`00449681`)
- New `services/config/sideeffects.Registry` — domain packages register their own per-key handlers; `api/base.go` just dispatches via `Registry.Dispatch`.
- `facilities.RegisterSettingsSideEffects` moves the inline checkout-toggle closure out of the API constructor.
- New `tenant.RegisterAfterCommit` so settings hooks can return a post-commit closure for non-transactional side effects (file unlinks, broadcasts) that must not run on rollback. Fires only after the outermost `WithTenantTx` commits successfully.
- `realtime.Broadcaster.BroadcastToTenant` promoted to the shared interface (already implemented on `Hub`); recipient counter and structured logging added.
- New `EventTenantSettingsChanged` event type. Settings and operator settings handlers fire it after every successful PUT/DELETE (4 call sites) so cross-origin tabs (operator ↔ tenant) invalidate their caches via SSE — same-origin tabs are already covered by the `phoenix:settings-changed` BroadcastChannel.
- `SettingsResource.OnValueSet` callback signature changes from `error` to `(postCommit func(), err error)`. Test mocks updated.

**Frontend** (`cda194c6`)
- Single SWR cache key (`SETTINGS_SCHEMA_SWR_KEY`) shared by `settings-page`, `sidebar`, `mobile-bottom-nav`, and `useTimetableDayHours`.
- New `useSettingsCacheBridge` hook, mounted once in the protected layout. Listens to two signals and invalidates the SWR cache on either:
  - same-origin `BroadcastChannel` (`phoenix:settings-changed`)
  - `phoenix:tenant-settings-stale` window event, dispatched by `useGlobalSSE` on `tenant_settings_changed`
- `SettingsContent` no longer owns local schema state, no `setTimeout` delayed refetches.
- `TenantProvider` continues to re-resolve its own tenant context on the broadcast (separate concern from schema invalidation) and tolerates out-of-order `resolveTenant` responses via a sequence-token guard.

## Out of scope

- The photo feature itself — follows in a stacked PR.
- `SupervisionContext.adminOverviewEnabled` cross-tab refresh (sourced from `/api/active/supervisors/all`, not the settings schema). Pre-existing behaviour, separate PR if we want it.

## Test plan

- [x] `go build ./...`, `tsc --noEmit`, `oxlint` green
- [x] Settings + Tenant-Provider unit tests: 33/33 green
- [x] Manual: cross-tab same-origin sync (Tab A change → Tab B reflects ~1–2 s without reload)
- [x] Manual: cross-origin operator → tenant sync via SSE
- [x] Manual: Schulhof + WC side-effect provisioning still fires on toggle
- [x] Manual: initial `/settings` load issues exactly one `/api/settings/schema` request

## Stats

33 files changed, +1208 / −261, 2 commits.

## Risk

Low. Behaviour-preserving refactor. Clean revert if needed.


yungweng (commented):
@theitger I reviewed the repeated review findings and checked the actual code paths. I think there are two real issues here, not three separate cache bugs.

## Summary
The backend-side architecture looks sound overall: the side-effect registry, after-commit hooks, and tenant-scoped SSE are the right direction. The frontend move to a single SWR read path for the settings schema is also reasonable because several components consume the same schema.

The problem is at the boundary between SWR freshness and the Next/RSC tenant-resolve cache. `operations.student_photos_enabled` travels through `/auth/tenant/resolve`, so refreshing `SETTINGS_SCHEMA_SWR_KEY` is not enough, and `router.refresh()` is not enough unless the `tenant-${slug}` cache tag has been purged first.

## Must fix
- `frontend/src/components/settings/settings-page.tsx:137`: `router.refresh()` can still reuse the cached tenant layout fetch from `frontend/src/app/[tenant]/layout.tsx:29`, which is tagged as `tenant-${slug}` with a 300s revalidate window. The tenant settings proxy route at `frontend/src/app/api/settings/values/[key]/route.ts` only forwards `PUT`/`DELETE`; it does not call `revalidateTag`, despite `settings-keys.ts` saying it does. This leaves `tenant.settings` stale for `operations.student_photos_enabled` after save/reset.

- The same cache invalidation gap exists for operator writes via `frontend/src/app/api/operator/provisioning/schools/[id]/settings/values/[key]/route.ts`. SSE can update already-open tenant tabs, but fresh reloads or clients that missed SSE can still receive stale `/auth/tenant/resolve` data until the 300s cache expires. Since this route only has `schoolId`, it needs a deliberate slug lookup or some equivalent way to purge the affected `tenant-${slug}` tag/path.

## Critical
- `frontend/src/components/tenant/tenant-provider.tsx:61`: resetting `requestSeqRef.current = 0` on every `tenantSlug` change can let an old in-flight resolve from the previous tenant pass the token check after a new tenant request reuses the same token. Keep the sequence monotonic across slug changes and/or compare the resolved slug before calling `setTenant`.

## Suggested shape
- For tenant-side settings writes, invalidate `revalidateTag(`tenant-${slug}`, { expire: 0 })` and likely `revalidatePath(`/${slug}`, "layout")` before relying on `router.refresh()`.
- For operator-side writes, resolve the affected school slug or call an existing revalidation endpoint with the right slug after successful `PUT`/`DELETE` for keys in `TENANT_RESOLVE_AFFECTING_KEYS`.
- In `TenantProvider`, do not reset the sequence counter to zero; also guard `setTenant` with `fresh.slug === requestedSlug`.

So: I like the general PR architecture, including SWR for the settings schema. But this needs the cache boundary fixed before merge, because `student_photos_enabled` is a privacy/visibility gate and stale values here are not acceptable.

yungweng (approved):
fick die katz, wird schon passne

PR #1403, opened 2026-05-12
feat: relocate student avatar and unify fallback colour

theitger:
## Summary
- `StudentCard` avatar moves from an absolutely-positioned bottom-right overlay to the top-left of the card header, inline with the name. The `useInlineHintAvatarRow` compact-surface workaround (and its OGS-groups call site) is dropped — the new layout works in every surface.
- Avatar initials-fallback background switches from the blue→green gradient (`#5080d8 → #83cd2d`) to a solid brand green (`#83CD2D`, `LOCATION_COLORS.GROUP_ROOM`) for both students and staff.
- Removed the top-left `animate-ping` (it pulsed through the new avatar position) and the pulse animation on the bottom-right deco ping; dropped the `pr-14` reserve that existed only for the old overlay.
- Backfilled `student-card` tests covering the photos-enabled / fallback-initials branches and the anmelden/abmelden tap-strip icons (new-code coverage ~31% → ~97%).

## Why
The bottom-right overlay collided with the "Tippen für mehr Infos" hint row and the tracking-indicator stack in dense lists — hence the `useInlineHintAvatarRow` branch. Moving the avatar into the header flow resolves it structurally instead of with a special case. The gradient fallback was a holdover that didn't match any brand colour; the solid brand green aligns with `LOCATION_COLORS`.

## Test plan
- [x] Student list with `photosEnabled = true`: avatar top-left, no overlap, ellipsis on long names — confirmed locally
- [x] Profile dropdown without photo: brand-green initials background — confirmed locally
- [ ] Student list without photos: green initials background
- [ ] Check-in flow: no stray arrow/hint
- [ ] `pnpm run check` green (passed in pre-push hook)

### DIFF
--- a/frontend/src/app/[tenant]/(protected)/ogs-groups/page.tsx
+++ b/frontend/src/app/[tenant]/(protected)/ogs-groups/page.tsx
@@ -1269,7 +1269,6 @@ function OGSGroupPageContent() {
                   onCheckinClick={() =>
                     void schoolCheckin.toggle(studentIdStr, checkinState)
                   }
-                  useInlineHintAvatarRow={photosEnabled}
                   locationBadge={
                     <StudentPresenceBadge
                       student={{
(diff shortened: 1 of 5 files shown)


yungweng (approved):
Ich vertrau

PR #1404, opened 2026-05-13
fix(api/common): guard error helpers against nil input (#423)

### LINKED ISSUE #423: [BUG] api/common error helpers panic on nil error
## Bug Description

The error helper functions in `api/common/errors.go` panic when passed a `nil` error because they call `err.Error()` without a nil check.

## Affected Functions (lines 38-126)
- `ErrorInvalidRequest(err error)`
- `ErrorUnauthorized(err error)`
- `ErrorForbidden(err error)`
- `ErrorNotFound(err error)`
- `ErrorInternalServer(err error)`
- `ErrorConflict(err error)`
- `ErrorBadGateway(err error)`
- `ErrorTooManyRequests(err error)`
- `ErrorGone(err error)`

## Root Cause

All error functions create an `ErrResponse` struct with:
```go
ErrorText: err.Error(),  // line 44 - panics if err is nil
```

## Impact

If any code path passes a `nil` error to these functions, the server will panic. While this is unlikely in normal operations, defensive programming suggests checking for nil.

## Suggested Fix

Add nil check before calling `err.Error()`:
```go
func ErrorInvalidRequest(err error) render.Renderer {
    errText := "unknown error"
    if err != nil {
        errText = err.Error()
    }
    return &ErrResponse{
        Err:            err,
        HTTPStatusCode: http.StatusBadRequest,
        Status:         "error",
        ErrorText:      errText,
    }
}
```

## Found By

Test file: `backend/api/auth/errors_test.go` - `TestErrorRenderersWithNilError`

## Reproduction

```go
common.ErrorInvalidRequest(nil) // panics: runtime error: invalid memory address or nil pointer dereference
```

theitger:
## Summary
- Route all 8 single-error helpers (`ErrorInvalidRequest`, `ErrorUnauthorized`, `ErrorForbidden`, `ErrorNotFound`, `ErrorInternalServer`, `ErrorConflict`, `ErrorTooManyRequests`, `ErrorGone`) plus `ErrorConflictWithCode` through a shared `newErrResponse(status, err)` constructor. When `err == nil`, it falls back to `http.StatusText(status)` and emits a `slog.Warn` so the caller bug surfaces instead of panicking with a nil-pointer dereference.
- `Err` stays `nil` on the response struct, so `RenderError`'s existing `errResp.Err != nil` guard keeps Sentry from capturing nil errors.
- Non-nil behaviour is bit-equivalent to before — pinned by a new regression test so a future "always use StatusText" refactor can't silently swallow real error messages.

Closes #423.

## Test plan
- [x] `go test -count=1 ./api/common/...` — all tests pass (incl. 3 new regression tests covering every helper for nil and non-nil inputs).
- [x] `go build ./...` clean.
- [x] `golangci-lint run ./api/common/...` — 0 issues.
- [x] `go test ./test/ -run TestHermeticTestPatterns` passes.
- [x] Full backend suite run against test DB — only pre-existing failures surface (`tenant_resolve_photos`, `TestRepairPickupScheduleTenantIDs`), verified via `git stash` baseline. Untouched by this PR.
- [x] Repo-wide grep confirms no production caller passes `nil` to any of these helpers today — the new `slog.Warn` branch is a defensive net, not a hot path.

### DIFF
--- a/backend/api/common/errors.go
+++ b/backend/api/common/errors.go
@@ -99,14 +99,30 @@ func (e *ErrResponse) Render(_ http.ResponseWriter, r *http.Request) error {
 	return nil
 }
 
-// ErrorInvalidRequest returns a 400 Bad Request error response
-func ErrorInvalidRequest(err error) render.Renderer {
+// newErrResponse builds an ErrResponse, tolerating a nil err by falling back to
+// the standard HTTP status text. A nil err here means a caller violated the
+// helper contract — we log a warning so the bug surfaces instead of being
+// masked by a panic or a silent "unknown error" string.
+func newErrResponse(status int, err error) *ErrResponse {
+	text := http.StatusText(status)
+	if err != nil {
+		text = err.Error()
+	} else {
+		slog.Default().Warn("error helper called with nil error",
+			slog.Int("status", status),
+		)
+	}
 	return &ErrResponse{
 		Err:            err,
-		HTTPStatusCode: http.StatusBadRequest,
+		HTTPStatusCode: status,
 		Status:         "error",
-		ErrorText:      err.Error(),
+		ErrorText:      text,
 	}
+}
+
+// ErrorInvalidRequest returns a 400 Bad Request error response
+func ErrorInvalidRequest(err error) render.Renderer {
+	return newErrResponse(http.StatusBadRequest, err)
 }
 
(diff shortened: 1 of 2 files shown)


yungweng (approved):
kuss digga, wird schon passne

PR #1436, opened 2026-05-18
fix: redirect binary caregivers to student search

fl0r14n28:
## Summary
- route caregivers in binary presence mode to `/students/search` after login
- pass tenant presence mode into SmartRedirect path selection
- add regression coverage for binary-mode caregiver redirects

## Verification
- `pnpm exec vitest run src/lib/redirect-utils.test.ts src/components/auth/smart-redirect.test.tsx --config vitest.config.ts`
- `pnpm run check`
- pre-push hooks: frontend knip, frontend check, go vet, go deadcode, golangci-lint


yungweng (approved):
ja diese gut

PR #1481, opened 2026-05-26
fix(students): start detail and history pages at top instead of inheriting list scroll

theitger:
## Problem

Closes #1461.

Next.js App Router keeps the current scroll position on navigation when the new page is already visible in the viewport. Scrolling down the student search list and clicking a child therefore opened the detail page mid-scroll, with the name/photo/status header off-screen. The same happened opening the **room-history** and **feedback-history** sub-pages from a scrolled detail page.

## Fix

- New reusable `useScrollToTop(key)` hook (`lib/hooks/use-scroll-to-top.ts`): `window.scrollTo(0, 0)` in an effect keyed on a stable id. Keyed (not bare `[]`) so it also fires when switching between students in place (no remount) — but **not** on unrelated re-renders.
- Applied to `students/[id]` (detail), `room-history`, and `feedback_history`. Replaces room-history's inline `useEffect` workaround.
- Unit test covers: mount fire, key-change re-fire, same-key no-op, no-key mount.

## Why it's safe for back/forward

The hook is keyed on the student id and Next's router cache preserves the page instance, so browser **back** still restores the previous scroll position (verified live: returning to a detail page restored to its prior offset, not 0). A layout/`template`-level handler keyed on `pathname` would *not* be safe — it would re-fire on back-navigation and clobber restoration.

## Verification

Driven live in the browser (demo-school, real client-side navigation):
- Search list scrolled to y=3000 → click child → detail page at **y=0**, header visible.
- Detail scrolled → room-history (client-side) → **y=0**.
- Detail scrolled → feedback-history (client-side) → **y=0**.
- Back-navigation into a detail page **restores** the prior scroll (no regression).

`pnpm run check` green (0 warnings / 0 errors); new hook test + existing room-history scroll test pass.

## Out of scope

Operator and parents portals have structurally similar full-page detail views reached from scrollable lists; left out of this PR by decision. Most tenant detail views use modals/slide-overs and aren't affected.

### DIFF
--- a/frontend/src/app/[tenant]/(protected)/students/[id]/feedback_history/page.tsx
+++ b/frontend/src/app/[tenant]/(protected)/students/[id]/feedback_history/page.tsx
@@ -7,6 +7,7 @@ import { Alert } from "~/components/ui/alert";
 import { useSession } from "next-auth/react";
 import { getStartDateForTimeRange } from "~/lib/date-helpers";
 import { useStudentHistoryBreadcrumb } from "~/lib/breadcrumb-context";
+import { useScrollToTop } from "~/lib/hooks/use-scroll-to-top";
 import { BackButton } from "~/components/ui/back-button";
 import { Loading } from "~/components/ui/loading";
 import { createLogger } from "~/lib/logger";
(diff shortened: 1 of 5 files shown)


yungweng (approved):
wird schon passen

PR #1515, opened 2026-06-02
fix: portal mobile filter panel so it clears the bottom nav

theitger:
## What

The mobile filter panel was being painted over by the bottom navigation, which blocked the "Anwenden" button.

Root cause: the panel renders inside the page content's `relative z-10` stacking context (`app-shell.tsx`), while the bottom nav is a `fixed z-30` element at the shell root. A `z-index` inside a nested stacking context can't paint above an element outside it — so the panel's `z-50` lost to the nav's `z-30`.

## Fix

- Portal the panel to `document.body` so it escapes the `z-10` context and its `z-50` competes at the root level (guarded with `typeof document` for SSR).
- Re-anchor the mobile sheet (`bottom-3` → `bottom-24`) so it sits above the bottom nav, tighten max height (`85vh` → `80vh`), and add `overscroll-contain` to stop scroll-chaining to the body.

## Notes

- Side effect (intended): the click-outside backdrop now also covers the header, so tapping the header while the panel is open dismisses it — standard popover behavior.
- Tests: existing suite updated to query the portaled DOM; 25/25 pass. `pnpm run check` clean.

## Testing

Verified on a mobile viewport — panel clears the nav and "Anwenden" is tappable.

### DIFF
--- a/frontend/src/components/ui/page-header/MobileFilterPanel.test.tsx
+++ b/frontend/src/components/ui/page-header/MobileFilterPanel.test.tsx
@@ -165,7 +165,7 @@ describe("MobileFilterPanel", () => {
     ];
 
     it("renders grid options in 2 columns", () => {
-      const { container } = render(
+      render(
         <MobileFilterPanel
           isOpen={true}
           onClose={mockOnClose}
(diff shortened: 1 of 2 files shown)


yungweng (approved):
lgtm, wird schon passen

PR #1517, opened 2026-06-03
feat: select existing guardian when creating/managing a student (#1513)

theitger:
## Why

A second child of the same parent forced staff to **re-type the guardian from scratch**, producing duplicate guardian profiles — duplicated contact data, split pickup/emergency permissions, and a messy data set for GDPR exports. This adds an **existing-guardian picker** to both the student-create flow and the student-detail guardian manager, so a sibling's guardian is *linked*, never duplicated.

## What changed

**Backend**
- New `GET /guardians/search` — a deliberately minimal, enumeration-resistant projection (name, email, and only a **count** of other linked children — never their names, addresses, notes, language, or account). Gated on `users:read`, same as the existing guardian list.
- `guardian_profile_id` accepted on the atomic student-create guardian input: when present the existing profile is linked as-is (never created, never mutated, no phone writes) — only the relationship flags apply to the new link.
- `LinkIfNotExists` repo method (idempotent `ON CONFLICT DO NOTHING` on `(tenant_id, student_id, guardian_profile_id)`): re-linking is a silent no-op, never a unique-violation that would abort the tenant transaction. Existing relationship flags are **never** overwritten by a re-link — flag changes stay on the dedicated update path.
- `CreateGuardian` now rejects a duplicate email as a `400` ValidationError ("…bereits vergeben – bitte die vorhandene Person über die Suche auswählen") instead of a raw `500`, steering the user to the picker.
- Linked-children enrichment batched in one query — no per-guardian N+1.

**Frontend**
- New inline `GuardianPickerPanel` (search → pick → set relationship → confirm). Inline rather than a second modal because a lookup is light; "Neu anlegen" still opens the full form.
- Shared `guardian-relationship-fields` extraction so the create form and the picker can't drift on relationship/permission UI.
- Wired into `student-create-modal` (collects into the atomic payload) and `student-guardian-manager` (links immediately).
- Client guardrails mirroring the server: 3-char minimum, 300 ms debounce, result-cap hint, greys out already-added guardians.

## Security / GDPR notes
- The picker exposes **strictly less** than the existing `users:read` `GET /guardians` list (which already returns full profiles), so it does not widen the data surface — it narrows it.
- Enumeration defenses: server-enforced 3-char minimum, LIKE-metacharacter escaping with explicit `ESCAPE '\'`, a hard 50-result cap (since `ParsePagination` has no upper bound), and a token-count cap to keep the WHERE clause bounded.
- Tenant isolation is RLS-enforced on the ambient tenant transaction throughout.

## Tests
- **Backend** (hermetic, DB-backed): picker permission gate (allowed for `users:read`, 403 without), GDPR projection (count not names), full-name token search, literal-wildcard search, short-query→empty, page_size clamp, repo-level `SearchByText` contract, `LinkIfNotExists` idempotency + tenant auto-set, select-existing create / duplicate-skip / mixed new+existing / not-found→400, and duplicate-email→ValidationError.
- **Frontend**: full picker select→confirm flow, flag pass-through, exclude/greying, error path, min-length guard, plus manager/modal link-existing wiring.
- Coverage on new code ~87% backend / ~98% picker component; remaining uncovered is defensive-only (nil guards, DB-error wrapping).
- `cd frontend && pnpm run check` clean; affected backend suites + hermetic check green.

## Reviewer call-out (one open item)
`CreateGuardian`'s new duplicate-email message ("…über die Suche auswählen") is also inherited by `CreateGuardianWithInvitation`, where there's no picker. Not a regression (that path failed before too, just with a 500), but if we want context-appropriate wording there it's a small follow-up.

## Cross-repo / migrations
None. No new env vars, no schema migrations (reuses the existing `idx_students_guardians_tenant` unique index), no IoT/PyrePortal contract changes.

### DIFF
--- a/backend/api/guardians/api.go
+++ b/backend/api/guardians/api.go
@@ -65,6 +65,12 @@ func (rs *Resource) Router() chi.Router {
 		// Guardian profile CRUD operations
 		// Read operations require users:read permission
 		r.With(authorize.RequiresPermission(permissions.UsersRead), withTx).Get("/", rs.listGuardians)
+		// Guardian picker search (#1513): same users:read gate as the rest of the
+		// guardian reads, so anyone who can reach the student create/detail flows
+		// can link a sibling's existing guardian. Returns a minimal,
+		// enumeration-resistant projection (name, email, linked-children count) —
+		// see searchGuardiansForPicker.
+		r.With(authorize.RequiresPermission(permissions.UsersRead), withTx).Get("/search", rs.searchGuardiansForPicker)
 		r.With(authorize.RequiresPermission(permissions.UsersRead), withTx).Get("/{id}", rs.getGuardian)
 		r.With(authorize.RequiresPermission(permissions.UsersRead), withTx).Get("/without-account", rs.listGuardiansWithoutAccount)
 		r.With(authorize.RequiresPermission(permissions.UsersRead), withTx).Get("/invitable", rs.listInvitableGuardians)
--- a/backend/api/guardians/guardians_search_test.go
+++ b/backend/api/guardians/guardians_search_test.go
@@ -1,0 +1,362 @@
+package guardians_test
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/url"
+	"testing"
+	"time"
+
+	"github.com/go-chi/chi/v5"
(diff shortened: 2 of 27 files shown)


yungweng (approved):
lgtm; wirklich guter PR. Das erste Mal wo 0 issues gefunden wurden. LFG!

PR #1528, opened 2026-06-03
Add scroll-to-error handling to student create modal

theitger:
## Summary

This PR adds scroll-to-error handling to the student create modal so failed submissions bring the first invalid field or submit error into view.

To avoid duplicating the enrollment form logic, the existing behavior is moved into a shared `useScrollToFirstError` hook and reused by both forms.

## Changes

- Adds `useScrollToFirstError` for long forms with inline validation errors.
- Wires the hook into `StudentCreateModal` and marks invalid student form fields with `aria-invalid`.
- Refactors the enrollment form to use the shared hook instead of local scroll state.
- Adds a regression test that verifies the student submit helper receives the error-scroll callback.

## Checks

- `cd frontend && pnpm run check`
- `cd frontend && pnpm test:run src/components/students/student-create-modal.test.tsx`
- Pre-push hook passed `frontend-knip` and `frontend-check`

### DIFF
--- a/frontend/src/components/enrollment/enrollment-form.tsx
+++ b/frontend/src/components/enrollment/enrollment-form.tsx
@@ -1,6 +1,6 @@
 "use client";
 
-import { useEffect, useMemo, useRef, useState } from "react";
+import { useEffect, useMemo, useState } from "react";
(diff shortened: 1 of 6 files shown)


yungweng (approved):
wird schon passen

PR #1533, opened 2026-06-04
Fix help guide header navigation

fl0r14n28:
## Summary

- add a reusable sticky help header for the guide overview and guide detail pages
- make the overview return button go back to the app via browser history with a tenant-root fallback
- keep guide detail pages focused with a `Zur Übersicht` link and no centered `moto Anleitung` label
- keep the PDF download button on one line

## Tests

- `pnpm vitest run src/components/help/guide-components.test.tsx src/proxy.test.ts`
- `pnpm run check`
- pre-push checks: `frontend knip`, `frontend check`, `go vet`, `go deadcode`, `golangci-lint`


yungweng (approved):
lgtm

PR #1600, opened 2026-06-09
Feat/bus kind ssot

christiank119:
## Summary

  Makes `users.students.bus_days` the **single source of truth**
  for the Buskind
  flag and removes the legacy `bus` boolean entirely. Follow-up
  to the foundation
  PR #1579, which introduced `bus_days` alongside `bus` as a
  synced compatibility
  field. This PR finishes the migration: `bus` is no longer
  written, read, or
  stored as an independent fact — it's derived from
  `bus_days.HasAny()` at the API
  boundary, and the DB column is dropped.

  Closes #1580

  ## What changed

  ### Backend
  - **Responses** derive the legacy `bus` flag from
  `bus_days.HasAny()` instead of
    reading the DB column.
  - **Write paths** (create/update + enrollment decision
  service) write only
    `bus_days`. The legacy `bus` boolean is still accepted as an
  input alias
    (`true` → Mo–Fr) but is ignored when `bus_days` is also
  supplied, and never
    flattens an existing per-day selection.
  - **Admin filter** `bus=yes/no` keeps its wording but
  evaluates `bus_days`
    semantics.
  - **Enrollment**: `student.bus_days` is the canonical reserved
  target;
    `student.bus` is kept as a legacy alias so older saved form
  schemas still
    resolve (and is hidden from the picker for new fields).
  - **Migration `1.15.116`**: defensively backfills any drifted
  rows, drops the
    `idx_students_bus` index, and drops the `users.students.bus`
  column. The
    `Bus *bool` model field and its SELECT column expressions
  are removed.
    Rollback re-adds the column and rebuilds it from `bus_days`.
  - **Seeder** sends `bus_days`.

  ### Frontend
  - Student mapping derives `bus` / `buskind` from `bus_days`;
  all submit paths
    send only `bus_days`.
  - Student detail cards and the data-management view show the
  exact weekdays
    (e.g. `Mo, Fr`) via `formatBusDays` instead of `Ja`/`Nein`.
  - The "Persönliche Informationen" edit modal now uses the
  shared per-day
    `BusStatusSection` weekday picker (previously a Ja/Nein
  dropdown), consistent
    with the create/edit modals.
  - Enrollment form editor offers `student.bus_days` as the
  Buskind target.

  ### CSV import
  - Legacy `Bus` column → all weekdays (Mo–Fr) when `Ja`.
  - New optional per-day columns `Bus.Mo`–`Bus.Fr` (override the
  legacy column
    when present), mirroring the existing `Ankunft`/`Abholung`
  per-day pattern.
  - Both CSV and XLSX templates now expose all six columns with
  example data
    demonstrating both modes.

  ## DB column decision

  The `bus` column is **dropped** (not kept as
  generated/read-only). There are no
  external consumers — PyrePortal/IoT never read it, and the
  JSON `bus` field is
  still served (derived) for API compatibility. The defensive
  backfill + rollback
  keep the migration safe.

  ## Acceptance criteria

  - [x] `bus_days` is the only field written by new frontend
  flows
  - [x] `bus` is never an independent source of truth in backend
  business logic
  - [x] `bus=yes/no` filters keep working with `bus_days`
  semantics
  ## Testing

  - Backend: `migrate reset` applies `1.15.116` cleanly (column
  + index dropped,
    `bus_days` retained); unit + DB integration tests pass for
  create/update
    (incl. legacy-alias preservation), repo persistence, admin
  filters/export,
    CSV import (legacy + per-day), enrollment targets, the
  drop-migration
    round-trip, and the hermetic-patterns CI check.
  - Frontend: `pnpm run check` (lint + typecheck) green; updated
  component/unit
    tests for the weekday display, `busDaysFromToggle`, and the
  per-day edit modal.


yungweng (approved):
lgtm

PR #1641, opened 2026-06-12
fix: wrap mobile master-detail list pane

theitger:
## Summary
- Wrap the mobile master-detail list pane in the same bordered white container treatment used elsewhere
- Set the drawer content background to white

## Notes
- This branch is cherry-picked directly from `development` so it contains only the UI tweak and no slot-list feature history.


yungweng (approved):
approved!

PR #1645, opened 2026-06-13
fix: batch SSE checkout events on session end (#848)

### LINKED ISSUE #848: SSE client channel buffer overflow on bulk checkout (events dropped)
## Context

Production log analysis 2026-02-05, 14:42 CET. 35 warnings in a short burst.

## Problem

During a bulk `student_checkout` operation, the SSE client channel for **user_id 26** (groups edu:5 + 237) filled up, causing **35 events to be dropped**. The server logs "SSE client channel full" when it cannot write to a client's channel buffer.

This means the connected client missed real-time updates during a critical operation (bulk checkout).

## Expected behavior

- SSE channel buffer should handle burst traffic from bulk operations
- If the buffer fills, the server should either increase capacity, drop oldest events (not newest), or notify the client to do a full refresh
- Clients should detect missed events and reconcile state

## Steps to investigate

1. Find the SSE server implementation — what is the current channel buffer size?
2. Determine why 35 events fired simultaneously (bulk checkout logic — does it send one event per student?)
3. Consider batching: send a single "bulk_checkout" event instead of N individual events
4. Alternatively, increase the channel buffer size for burst scenarios
5. Add client-side detection for missed events (sequence numbers or reconnect-and-refresh)

## Production log excerpt

```
"SSE client channel full" (35x, 14:42)
Event type: student_checkout
User: user_id 26, groups edu:5 + 237
```

theitger:
Fixes #848.

## Problem

On a whole-session checkout, `broadcastStudentCheckoutEvents` looped per student and emitted **2N+1** SSE events onto a single client's channel (one `student_checkout` to the active-group topic + one to the edu-group topic per student, plus a global dashboard refresh). The client channel buffer was only 10 slots, so any session with >5 students dropped events ("SSE client channel full", 35x in the production log that filed the issue).

## Fix (approaches A + B from the issue; C intentionally out of scope)

**B — batch the events (root cause):**
- New event type `bulk_student_checkout`. On session end we now emit **one event per topic**: one to the active-group topic carrying every student ID, and one per distinct edu group carrying that group's IDs. `2N+1` → `~3` events.
- Single-student checkout is unchanged (still `student_checkout` via `broadcastVisitCheckout`).
- The per-student attendance snapshot the old loop computed (at a cost of 2N schedule queries) is dropped — SSE events are triggers, not payloads; the client refetches, and the frontend never read those fields.

**A — buffer headroom:**
- Client channel buffer `10 → 32`.

**Frontend:**
- Handle `bulk_student_checkout` (mirrors `student_checkout` invalidation: group cache + each student's detail cache).
- Register the named-event listener in `use-sse.ts`; sync `sse-types.ts`.

**C (sequence numbers / client gap detection):** intentionally not implemented — over-engineering for this issue.

## Verification

- `go build ./...` clean; `go test ./services/active/` passes incl. new `TestBroadcast_EndActivitySessionBatchesCheckouts`.
- Frontend `pnpm run check` clean; 68 SSE hook tests pass (incl. new bulk-checkout coverage in `use-sse` + `use-global-sse`).

### DIFF
--- a/backend/api/sse/sse_helpers.go
+++ b/backend/api/sse/sse_helpers.go
@@ -199,7 +199,7 @@ func (conn *sseConnection) sendHeartbeat() error {
 // createAndRegisterClient creates the SSE client and registers it with the hub
 func (rs *Resource) createAndRegisterClient(conn *sseConnection) {
 	conn.client = &realtime.Client{
-		Channel:          make(chan realtime.Event, 10), // Buffer up to 10 events
+		Channel:          make(chan realtime.Event, 32), // Buffer up to 32 events (issue #848: headroom for bursts, e.g. admin-overview clients subscribed to every group)
 		UserID:           conn.staffID,
 		TenantID:         conn.tenantID,
 		SubscribedGroups: make(map[string]bool),
--- a/backend/realtime/events.go
+++ b/backend/realtime/events.go
@@ -14,6 +14,17 @@ const (
 	EventStudentCheckOut EventType = "student_checkout"
 	EventStudentUpdated  EventType = "student_updated"
 
+	// Bulk checkout event — emitted when a whole session ends (manual end or
+	// scheduler timeout) and every active visit is closed at once. One event
+	// per affected topic carrying all student IDs, instead of one
+	// student_checkout per student, so a single client receives one trigger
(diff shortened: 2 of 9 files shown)


yungweng (approved):
Image

PR #1646, opened 2026-06-13
fix: gate Schulhof status polling on groups:read permission (#846)

### LINKED ISSUE #846: Schulhof status polling floods logs with 403s (~150/day)
## Context

Production log analysis 2026-02-05, 10:00–15:00 CET. ~150 warnings in 5 hours.

## Problem

`GET /api/active/schulhof/status` is polled every ~60 seconds and consistently returns **403 Forbidden**. The requests originate from `server:8080` (internal Docker network), indicating this is **Next.js SSR** making backend calls with an expired or invalid JWT.

This is the **noisiest issue in the logs** — ~150 warnings in a 5-hour window — drowning out real problems.

## Expected behavior

- Server-side requests should use a valid JWT (refresh before expiry)
- If the token is expired, the SSR layer should refresh it before making the request
- If Schulhof status is not needed server-side, move the polling to the client

## Steps to investigate

1. Find the Next.js code that polls `/api/active/schulhof/status` — is it an API route, SSR fetch, or middleware?
2. Check how the JWT is obtained for server-side requests (is it from the session? A service token?)
3. Determine why the token is consistently invalid (expired? wrong audience? missing refresh?)
4. Either fix the token refresh flow or move this polling to client-side only
5. Consider reducing polling frequency or using SSE for real-time updates instead

## Production log excerpt

```
403 Forbidden — GET /api/active/schulhof/status (~150x)
Fires every ~60s from server:8080 (internal Docker)
Expired/invalid JWT
```

theitger:
Fixes #846.

## Problem
`SupervisionProvider` wraps the entire `[tenant]` layout and polls `GET /api/active/schulhof/status` every 60s for **every** authenticated tenant user. The backend gates that endpoint on `groups:read` (`api/active/api.go` → `RequiresPermission(permissions.GroupsRead)`), so accounts without it (e.g. the limited `guest` role) produced a guaranteed 403 on every poll — ~150 warnings per 5h in production, the noisiest log line. The client swallows the error (`.catch(() => null)`), so nothing broke in the UI; only the backend logs filled up.

## Fix
Stop the 403s at the source (issue's preferred option 1):
- Plumb the JWT `permissions` claim through token → session (`server/auth/shared.ts`).
- Port the backend authorizer to the client: `hasPermission()` in `auth-utils.ts` is a 1:1 mirror of `backend/auth/authorize/permission.go` (admin wildcards `admin:*`/`*:*`, resource/action wildcards, prefix wildcards), so the client gate can never diverge from what the backend allows.
- Skip the Schulhof fetch when the session lacks `groups:read`. Admins keep their role fallback; pre-rollout sessions without a `permissions` claim fall back to the caregiver role so they are not broken during the deploy window. The backend stays the final authority.

The other callers of the endpoint are not the flood source: `active-supervisions` page is an admin-only fallback path, the dashboard BFF route runs on a permission-gated page — neither is the "every user every 60s" poll.

## Tests
- `hasPermission` unit tests: exact match, missing, null session, admin wildcards, resource/prefix wildcards, malformed input.
- `supervision-context` integration: skips fetch without `groups:read`, still polls for admins and for legacy sessions without a permissions claim.
- `config` callbacks: permissions carried onto token and exposed on session.
- 218 tests pass; `tsc --noEmit` clean; `oxlint` 0 warnings.

## Notes
Intentionally the targeted fix, not the SSE migration also floated in the issue. Polling cadence is unchanged.

### DIFF
--- a/frontend/src/lib/auth-utils.test.ts
+++ b/frontend/src/lib/auth-utils.test.ts
@@ -3,6 +3,7 @@ import type { Session } from "next-auth";
 import {
   hasRole,
   hasAnyRole,
+  hasPermission,
   isAdmin,
   isCaregiver,
   isAuthenticated,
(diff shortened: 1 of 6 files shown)


yungweng (approved):
Image

PR #1647, opened 2026-06-13
fix: drop bot/scanner probe methods from request log (#850)

### LINKED ISSUE #850: Bot/scanner probes generate 405 log noise (CONNECT, HEAD, PRI)
## Context

Production log analysis 2026-02-05, 10:00–15:00 CET. 9 occurrences.

## Problem

External bots/scanners are probing the production server with non-standard HTTP methods:

- `CONNECT` — HTTP tunneling probe (common from proxies/scanners)
- `HEAD /` — Availability check (crawlers, uptime monitors, vulnerability scanners)
- `PRI *` — HTTP/2 prior knowledge probe

The backend correctly returns **405 Method Not Allowed**, but these still generate warning-level log entries that add noise.

## Current impact

- 9 warnings in 5 hours — low volume, but consistent
- Contributes to log noise alongside the higher-volume issues (#846, #847)
- No security risk (requests are rejected), but indicates the server is being probed

## Steps to investigate

1. **Log level**: Consider downgrading 405s from unrecognized methods to DEBUG instead of WARN — they carry no diagnostic value at warning level
2. **Reverse proxy filtering**: If Caddy/nginx sits in front, consider dropping CONNECT and PRI requests before they reach the Go backend
3. **Rate limiting**: Evaluate whether repeated probes from the same IP should trigger rate limiting or blocking
4. **Monitoring**: If these increase significantly, it could indicate targeted scanning — consider alerting on volume spikes rather than individual events

## Production log excerpt

```
405 Method Not Allowed (9x)
Methods: CONNECT, HEAD /, PRI *
Source: External bot/scanner probes
```

theitger:
## Summary

External bots/scanners probe production with HTTP methods our API never serves (`CONNECT` tunneling, `TRACE`/XST, `PRI` HTTP/2 preface). These get a 404/405 but add WARN-level log noise (~9/5h observed, issue #850).

This adds a second `slogchi` filter — `IgnoreMethod(CONNECT, TRACE, PRI)` — so those requests are no longer logged. Legitimate 4xx on normal methods (e.g. a `GET 404` on a bogus path) keep logging.

## Why drop instead of downgrade to DEBUG

These methods carry zero diagnostic value at any level. The Prometheus HTTP middleware runs **earlier** in the chain (`httpMetrics.Middleware` before the slog middleware) and still counts every request, so volume-based scan alerting is unaffected — only the per-event log lines disappear.

`HEAD /` is intentionally **not** touched: it resolves to the `GET /` handler and returns 200/INFO, so it never contributed to WARN noise.

## Scope

- Log-noise reduction only. Reverse-proxy (Caddy) filtering and per-IP probe rate-limiting from the issue's "steps to investigate" are infra concerns, left out of this code change.

## Test plan

- [x] `go test ./api/ -run TestLoggingMiddlewareIgnoresScannerMethods -v` — CONNECT/TRACE/PRI dropped, GET 404 still logged
- [x] `go build ./...`, `go vet ./api/`, `gofmt` clean

Closes #850

### DIFF
--- a/backend/api/base.go
+++ b/backend/api/base.go
@@ -176,6 +176,13 @@ func setupBasicMiddleware(router chi.Router, logger *slog.Logger, httpMetrics *o
 		WithTraceID:      false,
 		Filters: []slogchi.Filter{
 			slogchi.IgnorePath("/health"),
+			// Bot/scanner probes use HTTP methods our API never serves
+			// (CONNECT tunneling, TRACE/XST, PRI HTTP/2 preface). They get a
+			// 404/405 but add only WARN-level log noise (issue #850). Drop the
+			// log lines; the Prometheus HTTP middleware runs earlier in the
+			// chain and still counts them, so volume-based scan alerting is
+			// unaffected. Legitimate 4xx on GET/POST/... routes keep logging.
+			slogchi.IgnoreMethod(http.MethodConnect, http.MethodTrace, "PRI"),
 		},
 	}))
 	router.Use(middleware.Recoverer)
--- a/backend/api/base_logging_test.go
+++ b/backend/api/base_logging_test.go
@@ -1,0 +1,76 @@
+package api
+
+import (
+	"context"
+	"log/slog"
+	"net/http"
+	"net/http/httptest"
+	"sync"
+	"testing"
+
+	"github.com/go-chi/chi/v5"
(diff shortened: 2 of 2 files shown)


yungweng (approved):
Image

PR #1682, opened 2026-06-16
feat: parent-authored per-day pickup & arrival times

theitger:
## What

Lets guardians set a **one-day pickup and/or arrival override** for their child from the parents portal, gated by a new per-tenant setting `operations.parent_pickup_change_enabled` (**default on** — product decision; only the *time* is editable, not who may collect the child). Changes take effect immediately and broadcast a `student_updated` SSE event so supervisor views refetch.

## Why

Schools asked for parents to adjust a single day's collection time without phoning the team. Staff keep full control: any staff edit reclaims the day and the parents portal can no longer change it.

## Screenshots

**1. Parents portal — change the pickup time for a single day**
Guardians set a per-day pickup/arrival override. Shown here with an existing Wednesday override (11:11) already loaded and the hint that saving overwrites it.

1-parents-portal-abholzeit-modal

**2. Staff view — a parent-edited day in the care plan**
In the child detail (tab "Betreuungszeiten") the day modal opens with the blue notice "Diese Zeiten wurden von den Eltern über das Elternportal gesetzt" and warns that a staff entry replaces the guardian's input.

2-staff-tagesmodal-von-eltern

**3. Setting — `operations.parent_pickup_change_enabled`**
Per school under Einstellungen → Betrieb → Elternportal; enabled by default.

3-setting-abholzeit-toggle

## How

**Backend**
- **Migration 1.15.134**: adds `source` (`staff`|`guardian`) + `created_by_guardian` (FK `ON DELETE SET NULL`) to `schedule.student_{pickup,arrival}_exceptions`; `created_by` becomes nullable so a guardian-authored row is first-class. Deleting a guardian account clears the author link but **keeps the care instruction** (staff still rely on it). _(Bumped from 1.15.133 to avoid a version collision with the passkey-credentials migration that landed on `development`.)_
- Parent service `SubmitCareException` / `ListCareExceptions` / `DeleteCareException`: ownership check, feature gate, full-replacement per-leg semantics (a nil leg clears that leg), immediate write under a **per-child-day advisory lock**, broadcast on commit, today..+2 months window.
- Staff create/update/delete reclaim a guardian-authored day under the same lock, with a `staff_profile_required` guard; staff-on-staff races return 409.
- New endpoints under `/parent/me/children/{studentId}/care-exception` (GET/POST/DELETE).

**Frontend**
- `PickupTimeModal` in the parents portal; quick-action wired with a feature gate that keeps existing guardian overrides clearable even after a school opts out.
- Staff care-day modal flags **"Von Eltern"** days and confirms before overwriting a parent-set time; care-schedule cards show the marker.
- `ParentApiError` carries the backend error `code` so the German UI maps it to a localized message; i18n added in de/en/ru/sq.
- Help guide updated for the staff-side marker/overwrite flow.

## Tests
- Parent service: happy paths, staff conflicts (incl. staff owning the other leg), full-replacement/clear-leg, account-deletion survival, error propagation (DB errors surface, not swallowed), authorization (un-owned child).
- Staff handler conflict/reclaim paths; model author validation; advisory-lock tenant guard; frontend modal tests.
- New-code coverage ~**88%** statements on the SonarQube-counted files (≥85% gate).

## Notes for reviewers
- `parent_pickup_change_enabled` defaults **on** by design (confirmed with product). Frontend falls back to *off* only on a transient features-fetch failure.
- Cross-repo: no IoT endpoint/contract changes; PyrePortal unaffected.


yungweng (approved):
tl;dr approved

PR #1697, opened 2026-06-17
feat(enrollment): rename form-schema templates in place

theitger:
Closes #1696

## What

Adds the ability to **rename an enrollment form-schema template** without publishing a new version or touching its fields. A logical schema spans multiple version rows sharing one `name`; this renames every version row atomically so the version lineage keeps one shared name.

Two entry points in the admin editor:
- A standalone **„Umbenennen"** dialog in the template actions menu (`⋮`).
- An **editable name field in the builder header**. A name-only save renames in place (no redundant version); a name+content save renames *then* publishes one new version — in that retry-safe order (a failed publish never leaves a redundant version; a retried rename is a no-op).

## How

**Backend** (Handler → Service → Repository):
- `PATCH /enrollment/schema/{id}` (`config:manage`) → `renameSchema` handler.
- `FormSchemaService.RenameSchema`: trims/validates, no-ops on same name, pre-checks collisions via `ExistsByName`, and translates a 23505 race on the `(tenant_id, name, version)` unique index into `ErrFormSchemaNameExists` → `409` with stable code `enrollment.schema_name_exists`.
- Repo `ExistsByName` + `RenameByName`, both tenant-scoped.

**Frontend**:
- `renameSchema` API client + Next.js PATCH proxy route.
- Mapped `enrollment.schema_name_exists` → a German message so a name collision shows *why*, not a generic failure (this code had **no mapping** before — a real UX gap).
- Help guide updated to document renaming.

## Tests

Existing happy-path/collision/no-op/empty coverage, plus added coverage for **every error branch** to clear the 85% new-code threshold:
- **Service**: non-positive id, not-found, load fault (not flattened to 404), exists-check fault, rename-exec fault.
- **Proxy route**: invalid id → 400, upstream throw → 500.
- **API client**: PATCH body shape, id URL-encoding, 409 code→German translation, generic fallback.

The `isUniqueViolationOn` 23505 race fallback is intentionally left to the repository-level DB tests (fabricating a `pgdriver.Error` in a unit test isn't practical) — noted in the test comments.

### Verification
- Backend: service / handler / repository rename tests green against the test DB.
- Frontend: 71/71 across the three touched test files; `pnpm run check` clean.

## Notes for review
- Rename + field-publish are two requests, not one transaction — deliberate and documented inline (`enrollment-form-editor.tsx`).
- Per-version rename keeps phase bindings (which pin schemas by id) intact.

### DIFF
--- a/backend/api/enrollment/api.go
+++ b/backend/api/enrollment/api.go
@@ -110,6 +110,7 @@ func (rs *Resource) Router() chi.Router {
 			r.With(authorize.RequiresPermission("config:read")).Get("/{id}", rs.getSchemaByID)
 			r.With(authorize.RequiresPermission("config:manage")).Post("/", rs.publishSchema)
 			r.With(authorize.RequiresPermission("config:manage")).Put("/{id}", rs.updateSchema)
+			r.With(authorize.RequiresPermission("config:manage")).Patch("/{id}", rs.renameSchema)
 			r.With(authorize.RequiresPermission("config:manage")).Delete("/{id}", rs.deleteSchema)
 		})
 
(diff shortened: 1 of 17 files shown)


yungweng (approved):
approved

PR #1698, opened 2026-06-17
feat: fixed selectable pickup times for enrollment forms

theitger:
Closes #1691

## What

Admins can attach a fixed list of `HH:MM` times to the pickup-times (`schedule.pickup`) weekday-schedule field of an enrollment form. When set, the public and parents-portal form renders a **per-weekday dropdown limited to those times** instead of a free time input. An empty/absent list preserves the historical free-entry behaviour, so existing forms are unaffected.

## How

Enforced end-to-end as defense-in-depth — the dropdown restricts input, and the server independently re-checks every *visible* schedule answer so a scripted or stale client cannot POST an off-list time:

- **Model** (`form_schema.go`): `AllowedTimes` on `FormField`, validated at schema-publish time (HH:MM, unique, trimmed, pickup-target-only; rejected on info/arrival/other fields). New `WeekdaySchedule.ValidateAllowed`.
- **Service** (`form_visibility.go`, `request_service.go`): `validateConstrainedSchedules` runs in **both** `Submit` and `ReplaceEditable`, returning `ErrPickupTimeNotAllowed` (wraps `ErrInvalidSubmission`). Hidden fields are skipped to match the existing sanitize behaviour.
- **Handlers** (public + parent): map the sentinel to a stable **400** with code `enrollment.pickup_time_not_allowed`.
- **Frontend**: constrained dropdown via `CustomSelect`; a saved-but-now-off-list time stays visible flagged "nicht mehr verfügbar"; an off-list rejection marks the offending schedule field (not just the banner); German message in `enrollment-error-messages`.
- **i18n** (de/en/ru/sq) + **in-app help guide** updated (per project rules).

## Tests

Unit / integration / regression coverage for every previously-uncovered new-code path:
- Model: `allowed_times` validation incl. info-field rejection; `ValidateAllowed`.
- Handlers (public + parent): sentinel and wrapped-error → 400 + stable code.
- Service (DB-backed): `Submit` accepts allowed times + persists; **rejects off-list** (the gate); `ReplaceEditable` enforces the same on the edit path.
- Frontend: dropdown limited to configured times; free-entry fallback; stale-value flagged; off-list rejection marks the field.

## Verification

- Browser-verified end-to-end on `demo-school.localhost`: constrained dropdown renders `14:45`/`16:00`; happy submit → 201 (persisted `{"mon":"14:45","wed":"16:00"}`); off-list submit → 400 `enrollment.pickup_time_not_allowed`.
- Backend: `gofmt`/`go vet` clean, touched packages + `TestHermeticTestPatterns` green.
- Frontend: `pnpm run check` clean.

## Note (not a blocker)

A malformed time (e.g. `99:99`) from a scripted client also returns the `pickup_time_not_allowed` code with a "must be HH:MM" message, because the service wraps every `ValidateAllowed` error in that one sentinel. Harmless and unreachable through the UI; can split into a distinct code if preferred.


yungweng (approved):
Image

PR #1699, opened 2026-06-17
fix(operator): restore no-slash GET /auth/passkeys route

theitger:
## Where this stems from

Surfaced in the code review of **#1697** (enrollment form rename). The reviewer flagged it as an unrelated, pre-existing regression, so it is split out here rather than bundled into that PR.

The actual regression was introduced by **#1690 — "keep operator passkey login public"** (commit `d0f5f2c75`). That PR replaced

```go
r.Route("/auth/passkeys", func(r chi.Router) { r.Get("/", rs.PasskeyList) })
```

with an exact leaf

```go
r.Get("/auth/passkeys/", rs.PasskeyList)
```

to stop the protected `/auth/passkeys` subtree from shadowing the **public** `/auth/passkeys/login/*` routes. That fixed the shadowing, but narrowed the list endpoint to the **trailing-slash form only**.

## The bug

This router has no `RedirectSlashes` middleware, so chi treats `/auth/passkeys` and `/auth/passkeys/` as **distinct patterns**. The old `r.Route(...).Get("/")` form answered both; the new exact leaf answers only the slash form.

- The frontend proxy sends the slash form, so the UI is unaffected.
- **Direct authenticated operator clients** hitting `GET /auth/passkeys` (no slash) now get a **404**.

## The fix

Register the list endpoint under **both** forms, still as individual leaves so the public login routes stay reachable (no protected subtree on `/auth/passkeys`):

```go
r.Get("/auth/passkeys", rs.PasskeyList)
r.Get("/auth/passkeys/", rs.PasskeyList)
```

## Test

Added `TestOperatorPasskeyListRouteAcceptsBothSlashForms`: drives the real `Router()` and asserts both forms reach the auth middleware (**401, not 404**) — an unregistered path 404s before middleware runs, a registered protected path 401s without a token. Pins the compatibility detail so it can't silently regress again.

- `go test ./api/operator/...` green.

Draft until #1697 lands / reviewer confirms scope.

### DIFF
--- a/backend/api/operator/api.go
+++ b/backend/api/operator/api.go
@@ -207,6 +207,14 @@ func (rs *Resource) Router() chi.Router {
 		// because public passkey login also lives under /auth/passkeys/login/*.
 		// A protected subtree on /auth/passkeys shadows those public login
 		// routes in chi and makes anonymous passkey login fail with 401.
+		//
+		// Register the list endpoint under BOTH the no-slash and trailing-slash
+		// forms: chi treats "/auth/passkeys" and "/auth/passkeys/" as distinct
+		// patterns (no RedirectSlashes on this router), and the old
+		// r.Route("/auth/passkeys").Get("/") form answered both. The proxy
+		// sends the slash form, but direct authenticated operator clients hit
+		// the no-slash form, so dropping it 404s them.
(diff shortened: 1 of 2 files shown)


yungweng (approved):
lgtm

PR #1702, opened 2026-06-17
fix(settings): make Standard badge value-based, not override-based (#1680)

theitger:
## What

Fixes #1680. The "Standard" badge on settings now reflects whether the **value equals the registry default**, instead of merely whether a tenant override row exists.

## Why

Boolean toggles have no reset button. Once a tenant flipped a toggle, an override row existed in `config.setting_values` forever, so the badge reported "not default" permanently, even after the admin toggled back to the default value. The badge being override-existence-based was itself the bug.

## Change

- **Backend** (`schema_builder.go`): `isDefault := !hasOverride || jsonValuesEqual(value, def.Default)`. `jsonValuesEqual` compares by JSON representation, which handles the JSONB roundtrip (override values unmarshal as `float64`, registry defaults are typically `int`). The check only runs when an override exists (short-circuit), so the schema build cost is unchanged in the common path.
- **Frontend** (`settings-api.ts`): the optimistic schema update computes `is_default` by comparing the new value to `item.default` instead of hardcoding `false`, so the badge reappears immediately without waiting for a refetch.

## Design note: display-only by intent

This deliberately does **not** delete the override row when the value equals the default. That "clean" approach was considered and **rejected as a breaking change**:

- It would flip `HasTenantOverride()` to `false`, causing env-fallback consumers to silently fall through to the env var instead of the registry default.
- It would silently migrate admins who deliberately set a value matching today's default off that value if the default later changed (loses explicit intent).
- It would change persisted state + write spurious `reset` rows to `config.setting_audit`, with a blast radius across every settings consumer (scheduler, device auth, IoT checkin) rather than one UI badge.

Keeping the override row preserves explicit admin intent and the `HasTenantOverride` env-fallback contract. The fix is contained to the read/display path.

## Tests

- Backend: 4 new cases in `settings_service_test.go` (no override; boolean override == default; boolean override != default; number JSONB float64-vs-int roundtrip). Green locally.
- Frontend: new `applyOptimisticSchemaUpdate` case asserting the badge restores when a value returns to the default. Full suite green (31 passed).

## CI note

Local pre-commit/pre-push `oxlint` hooks fail on a stale `node_modules` (installed `oxlint` 1.62.0 vs the `^1.69.0` pin in `package.json`); the config references a11y rules that only exist in 1.69.0. This is environment drift unrelated to the diff (this change is plain TS, no JSX) and CI re-lints with the pinned version.

### DIFF
--- a/backend/services/config/schema_builder.go
+++ b/backend/services/config/schema_builder.go
@@ -60,9 +60,14 @@ func buildSchemaWithScope(
 			continue
 		}
 
-		// Determine if this is using the default (no tenant DB override exists)
+		// Determine if this is using the default. A setting is "default" when no
+		// tenant override exists OR the override happens to equal the registry
+		// default. The value-based check matters for boolean toggles: they have
+		// no reset button, so toggling back to the default value would otherwise
+		// leave is_default false forever (issue #1680). jsonValuesEqual handles
+		// the JSONB roundtrip (float64 vs int) and type-correct comparison.
 		hasOverride, _ := svc.HasTenantOverride(ctx, key)
-		isDefault := !hasOverride
+		isDefault := !hasOverride || jsonValuesEqual(value, def.Default)
 
 		// Mask password values (only when actually set, not empty defaults)
 		displayValue := value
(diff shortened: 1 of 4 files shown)


yungweng (approved):
wird schon

PR #1705, opened 2026-06-17
feat(enrollment): permanent same-ways-home toggle for all care days

theitger:
Closes #1695

## What & why

Currently the enrollment form lets parents pick the allowed ways home (Gehweise) **per care day**. Issue #1695 asks for the common case: choose the ways home **once** and have it apply to every day ("permanente Gehweise").

This adds a checkbox above the weekday grid in the multi-mode departure input that flips it between the existing per-day view and a single shared (uniform) selection. As the issue suggested, it reuses the existing endpoint: the shared selection is simply written to every care day in the payload.

## Behavior

- **Checkbox "Gleiche Heimwege für alle Betreuungstage"** toggles uniform mode. Default is the per-day view.
- **Enabling over identical data is lossless** — if every care day already holds the same selection, it is adopted as the shared selection and the view just flips.
- **Enabling over divergent per-day choices prompts a confirmation** ("Löschen und fortfahren") before discarding them, then the parent makes one fresh selection that fans out to all days.
- **Stays in sync with the care-day set** — an effect re-fans the shared selection when care days are added/removed elsewhere on the form (incl. a full swap), so a newly added day never silently lacks coverage.
- **Backend/validation/pruning untouched** — the stored value remains per-day (each day gets the same array); only the UI changes.

## Incidental fix

`ChildDraft` now carries a stable `clientId` used for the React list key instead of the array index. Without it, removing a child slot from the middle of the list leaked a card's local view state (uniform toggle, expanded days) onto a different child. Saved edit-draft children derive the key from the persisted child id so cards don't remount when offerings load.

## Tests

Added 11 tests covering: applying/deselecting a uniform mode across all days, the confirm/cancel flow, lossless adopt, the add-day and full-swap sync paths, the "pick care days first" prompt, locale-label fallback, edit-draft prefill, and per-child state isolation on sibling removal. `pnpm run check` passes; the enrollment-form suite is green (37 tests).

## i18n

New keys added to de/en/ru/sq.

## Scope note

This is the public parents-facing enrollment form, so the in-app help guide (staff app) is not affected.

### DIFF
--- a/frontend/src/components/enrollment/enrollment-form.test.tsx
+++ b/frontend/src/components/enrollment/enrollment-form.test.tsx
@@ -55,7 +55,23 @@ vi.mock("next-intl", async () => {
     string,
     unknown
   >;
-  const catalogs: Record<string, Record<string, unknown>> = { de, en };
+  // A German catalog with the weekday/departure label maps removed, exposed as
+  // a pseudo-locale. Selecting it via mockIntlLocale lets a test render with
+  // missing labels so the component's `labels[key] ?? key` fallbacks show the
+  // raw keys. Uses the same locale-switching path as the real "en" catalog.
+  const deNoLabels = structuredClone(de) as Record<string, unknown>;
+  const deNoLabelsEnrollment = deNoLabels.enrollmentForm as Record<
+    string,
+    unknown
+  >;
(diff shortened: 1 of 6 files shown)


yungweng (approved):
Image

PR #1740, opened 2026-06-22
fix(enrollment): align pickup schedule hints with validation gate

theitger:
## What

Fixes misleading hint copy on the enrollment `weekday_schedule` (pickup time) field so it matches what validation actually enforces.

The old inline hint branched only on `careConstrained`, ignoring `field.required`. As a result a **required but unconstrained** schedule rendered the optional copy *"Leere Felder bedeuten: an diesem Tag keine Angabe"* — telling parents blanks are fine, even though an all-empty schedule is rejected on submit.

## Change

`WeekdayScheduleInput` now branches the hint on `required` first, then `careConstrained`, matching the `customValueMissing` / `requiredMessageForField` gates exactly:

| State | Validation gate | Inline hint |
|---|---|---|
| not required | skipped | `emptySchedule` (blanks ok) |
| required + care-constrained | time on every care day | `everyDayTime` |
| required + unconstrained | at least one time | `atLeastOneTime` (new) |

Also:
- Adds `structured.atLeastOneTime` across all four locales (de/en/ru/sq).
- Dedupes `everyDayTime` / `scheduleEveryDay` to the same "für jeden Betreuungstag" wording (they describe the same constraint, shown inline + in the summary banner).
- Editor (`enrollment-form-editor.tsx`): required-state hint copy + a `student.pickup_status` branch mirroring the runtime rule (pickup accepts an empty, touched selection).
- Fixes German closing quotes (`„…"` → `„…"`) in the `dayMissing` error.

## Note for reviewers

One existing test assertion changed **behavior**, not just a string: the required+unconstrained case now expects `atLeastOneTime` instead of the old `emptySchedule` copy. This is the intended fix, not a test edited to go green. A new test (`keeps an optional pickup schedule lenient even when offerings constrain the form`) covers the previously-untested optional+constrained path.

## Verification

- `pnpm run check` (verify-locales + oxlint + tsc) — passes
- `enrollment-form.test.tsx` — 57/57 pass


yungweng (approved):
lgtm; kann kein deutsch

PR #1743, opened 2026-06-22
feat(parent): guardian contact & pickup management in the parents portal (#1667)

theitger:
## What this does
Lets eligible guardians manage, from the parents portal, the contact data and per-child pickup/emergency arrangements of the people around their child — replacing the disabled "Abholberechtigte Personen" placeholder with a working panel.

A parent can now:
- **See** every guardian of their child with contact + pickup detail.
- **Edit the contact data** (name, email, address, phone list) of their own profile and of contact-only helpers without their own login.
- **Set the per-child `can_pickup` / `is_emergency_contact` flags** and a pickup note for those helpers.

## Authorization model (the core of this PR)
Relationship-scoped, per `.claude/rules/guardian-parent-permissions.md`. Two new guardian permissions, granted only to the full guardian presets (primary/legal/co): `parent_portal.guardian.edit` and `parent_portal.pickup.manage`.

Editing is blocked — in the service, not just the UI — for any target that is:
- a **guardian with their own portal account** (they manage their own data),
- a **social worker** (school-managed; contact also redacted on read),
- a **full guardian without an account** (a real legal guardian, not a helper),
- a **profile shared with a child outside the caller's family**.

The safety-critical pickup/emergency flags can only ever be set on account-less helpers, and every change is written to an append-only audit trail (`audit.guardian_changes`). Read redaction reuses the same predicate as the write guard, so the listing never shows data a caller could then blank-overwrite.

## Backend
- New service `services/parent/parent_guardian_service.go` (+ handlers under `api/parent/`) exposing `GET /parent/me/children/{id}/guardians` and `PUT .../guardians/{gid}/{contact,pickup}`.
- New audit table `audit.guardian_changes` (tenant-RLS, append-only). Contact-field changes log *which* field changed, never the before/after PII values; pickup-flag changes log the boolean transition. Actor name/email snapshotted so the trail survives account deletion.
- Batch repo methods (`GuardianProfile.FindByIDs`, `GuardianPhone.FindByGuardianIDs`) to avoid N+1 when listing.
- Row-locking on the guardian profile serializes concurrent contact/relationship edits.
- A contact edit is shared across siblings; the SSE broadcast fans out to every child the profile serves, not just the routed one.

## Migrations
- `1.15.143` — backfills the two new permissions onto existing full-guardian relationships (idempotent, COALESCE-guarded).
- `1.15.144` — creates `audit.guardian_changes` with tenant isolation + indexes.

## Settings
- New `operations.parent_guardian_management_enabled` toggle (defaults **ON**, `config:manage`). The toggle only governs whether the feature is exposed; the pickup-authority safety is structural (helpers-only + audited), not toggle-based. Switching it off also disables a parent editing their **own** contact data — the description says so explicitly.

## Frontend
- New `GuardiansPanel` (parents portal child detail, desktop + mobile), built from the UI kit (`Modal`/`Input`/`Button`/`Checkbox`, `LOCATION_COLORS`). Capability flags from the API drive which edit affordances render, so the UI never offers a control the backend would reject.
- Locale strings added in all four languages (de/en/ru/sq); the removed placeholder's orphaned keys cleaned up.

## Testing
- Backend service tests cover every authorization boundary (account holder, social worker, full guardian, cross-family, helper-allowed), the pickup-vs-edit permission split, audit-row writes, feature-disabled, and read redaction.
- Frontend panel tests for primary-phone preservation, error-code mapping, and the redacted-affordance case.
- Verified locally: `go test ./services/parent/... ./api/parent/... ./database/repositories/{users,audit}/... ./services/config/...` green; `pnpm run check` + `knip` green; handler-layer + service-repository ratchets and the settings registry pass.

## Notes for reviewers
- `GuardianChange.ListByStudentID` is wired but has no production consumer yet — it's the read side for the planned admin per-child audit view (#1455).
- Help guide intentionally untouched: parents-portal-only changes are exempt per the help-guide-sync rule.

Refs #1667


yungweng (approved):
Image

Spaaaaaß, gerne mache ich review :)

PR #1762, opened 2026-06-25
feat(auth): parent password reset flow

theitger:
Closes #1758

## What

Self-service password reset for the **parents portal**: a "forgot password" request modal on the parents login, a tokenized reset page, and the backend flow scoped to guardian accounts (`InitiateParentPasswordReset`). Reset tokens are shared with the staff flow; dedicated parent proxy routes keep parent pages off the tenant auth API surface.

The feature commit also propagates role-lookup errors in `findGuardianTenantForAccount` (instead of swallowing them) so a real guardian never gets a neutral `200` with no token/email on an RLS/config/transient DB failure, and switches the reset email to the formal *Sie* address.

## Tests (this is primarily a coverage PR for the flow)

Brings new-code coverage above the 85% gate with regression-oriented unit + integration tests:

**Frontend**
- Parent `password-reset` + `confirm` proxy route handlers — success, `Retry-After` forwarding, JSON/text/empty/unparseable error bodies, 500 on upstream throw (100% lines)
- `ResetPasswordPageContent` — every `validatePassword` branch, status→copy mapping (410/404/400/generic), success + redirect, visibility toggles, tenant branding, default copy (100% lines, no test existed before)
- `parents/reset-password` wrapper page wiring
- `confirmParentPasswordReset` error path in `auth-api`

**Backend**
- Parent auth handlers — bind `400`s and generic `500`s for login/reset/confirm, plus invalid-token and weak-password `400` mapping
- `findGuardianTenantForAccount` — dangling `account_role` row skipped neutrally (data-inconsistency tolerance)

Each test documents the invariant it protects (account-enumeration masking, the `Retry-After` cross-layer contract, 400-vs-500 status semantics).

## Verification
- `pnpm run check` (verify-locales + oxlint + tsc): clean
- `go vet` / `golangci-lint` / knip: clean (all pre-push gates passed)
- New tests pass

## Notes
- A pre-existing, unrelated test (`TestNotesEndpoint_AddAndList`, parent-messaging #1672) fails locally only because that feature's migration isn't applied to the local DB; it passes in CI's fresh-migrate environment and is untouched here.
- No existing tests were modified to make new code pass.


yungweng (approved):
cricket cross

PR #2199, opened 2026-08-09
fix: use production favicon for parents app

fl0r14n28:
## What changed

The production parents hosts now resolve to the standard production favicon variant used by the OGS app. The legacy `parents.` hostname follows the same behavior. Parent staging hosts keep their staging marker.

## Why

The colored favicon markers are intended for internal surfaces such as the operator portal and staging environments. The production parents app is a public product surface and should use the normal production branding.

## Impact

The parents app will display the same moto favicon as the production OGS app. No image assets or other portal behavior changed.

## Root cause

The favicon resolver mapped production parents hosts to the marked `eltern` variant instead of the standard `normal` variant.

## Checks

- `pnpm exec vitest run src/lib/favicon-variants.test.ts`, 14 tests passed
- Pre-commit frontend lint and formatting hooks passed
- Pre-push frontend check, dependency cruise, Knip, Go vet, golangci-lint, deadcode, and git-secrets hooks passed


yungweng (approved):
No blockers or critical findings found.

